PR #2, opened 2015-09-24
WIP: Support for initialising the sandbox and git submodules

jacobstanley:
Mainly want to get feedback on the approach

Still needs some tests

### DIFF
--- a/mafia.cabal
+++ b/mafia.cabal
@@ -15,6 +15,7 @@ library
                       base                            >= 3          && < 5
                     , ambiata-p
                     , ambiata-x-either
+                    , async                           == 2.0.*
                     , bytestring                      == 0.10.*
                     , containers                      == 0.5.*
                     , directory                       == 1.2.*
--- a/mafia.cabal
+++ b/mafia.cabal
@@ -52,6 +52,7 @@ executable mafia
                     , ambiata-p
                     , ambiata-x-either
                     , ambiata-x-optparse
+                    , containers                      == 0.5.*
                     , directory                       == 1.2.*
                     , filepath                        == 1.4.*
                     , optparse-applicative            == 0.11.*
--- a/mafia.cabal
+++ b/mafia.cabal
@@ -14,8 +14,16 @@ library
   build-depends:
                       base                            >= 3          && < 5
                     , ambiata-p
-                    , either                          >= 4.4.1      && < 4.5
+                    , ambiata-x-either
+                    , bytestring                      == 0.10.*
+                    , containers                      == 0.5.*
+                    , directory                       == 1.2.*
+                    , exceptions                      == 0.8.*
+                    , filepath                        == 1.4.*
+                    , process                         == 1.2.*
                     , text                            == 1.2.*
+                    , transformers                    == 0.4.*
+                    , unix                            == 2.7.*
 
   ghc-options:
                     -Wall
@@ -27,6 +35,9 @@ library
                     Paths_ambiata_mafia
                     BuildInfo_ambiata_mafia
                     Mafia
+                    Mafia.IO
+                    Mafia.Path
+                    Mafia.Process
 
 executable mafia
   ghc-options:
@@ -39,9 +50,13 @@ executable mafia
                       base
                     , ambiata-mafia
                     , ambiata-p
+                    , ambiata-x-either
                     , ambiata-x-optparse
+                    , directory                       == 1.2.*
+                    , filepath                        == 1.4.*
                     , optparse-applicative            == 0.11.*
-                    , text
+                    , text                            == 1.2.*
+                    , transformers                    == 0.4.*
 
 
 test-suite test
--- a/mafia.cabal
+++ b/mafia.cabal
@@ -8,7 +8,7 @@ synopsis:              mafia
 category:              System
 cabal-version:         >= 1.8
 build-type:            Custom
-description:           project
+description:           mafia
 
 library
   build-depends:
@@ -26,7 +26,7 @@ library
   exposed-modules:
                     Paths_ambiata_mafia
                     BuildInfo_ambiata_mafia
-                    Project
+                    Mafia
 
 executable mafia
   ghc-options:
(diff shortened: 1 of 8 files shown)


src/Mafia/Process.hs
```diff
@@ -0,0 +1,268 @@
...
+  { processCommand     :: File
+  , processArguments   :: [Argument]
+  , processDirectory   :: Maybe Directory
+  , processEnvironment :: Maybe (Map EnvKey EnvValue)
+  } deriving (Eq, Ord, Show)
+
+------------------------------------------------------------------------
+
+-- | Pass @stdout@ and @stderr@ through to the console.
+data Pass = Pass
```
charleso, line 77: :+1:

src/Mafia/Path.hs
```diff
@@ -0,0 +1,39 @@
...
+
+type Path      = Text
+type File      = Path
+type Directory = Path
+
+------------------------------------------------------------------------
+-- Operations
+
+(</>) :: Path -> Path -> Path
+(</>) x y = T.pack (FilePath.combine (T.unpack x) (T.unpack y))
```
charleso: Hmm. This combinator is a little overloaded already in other places. That said, I guess `mafia` isn't going to be used as a library.

jacobstanley: It does mean the same thing as `System.FilePath.</>` just over the mafia `Text` based paths

jacobstanley: I updated the code to make it obvious that it's just relaying the call to the same operator in `System.FilePath`

mafia.cabal
```diff
@@ -40,8 +52,12 @@ executable mafia
                     , ambiata-mafia
                     , ambiata-p
                     , ambiata-x-optparse
+                    , directory                       == 1.2.*
+                    , either                          >= 4.4.1      && < 4.5
```
nhibberd: This bound could possibly cause issues (not sure if this project will be depended on up stream), a lot of our projects currently have a strict version of `== 4.3.*`.

jacobstanley: All good this was the bound from the anatomy template, maybe we should change that too?

src/Mafia/Process.hs
```diff
@@ -0,0 +1,271 @@
...
+  | ProcessException Process SomeException
+  deriving (Show)
+
+-- The `show x == show y` stuff is because SomeException doesn't supply an Eq
+-- instance and we would really like to be able to compare errors for testing
+-- purposes.
+instance Eq ProcessError where
+  (==) (ProcessFailure   p x) (ProcessFailure   q y) = p == q && x == y
+  (==) (ProcessException p x) (ProcessException q y) = p == q && show x == show y
+  (==) _                      _                      = True
```
EduardSergeev: Is it safe? If an additional constructor is added to `ProcessError`?

src/Mafia/Process.hs
```diff
@@ -0,0 +1,271 @@
...
+type ExitStatus = Int
+
+data ProcessError
+  = ProcessFailure   Process ExitStatus
+  | ProcessException Process SomeException
+  deriving (Show)
+
+-- The `show x == show y` stuff is because SomeException doesn't supply an Eq
+-- instance and we would really like to be able to compare errors for testing
+-- purposes.
```
EduardSergeev: If it is only for testing maybe it should be defined in `Test` module?

mafia.cabal
EduardSergeev: Is it even necessary to reference `either`? `ambiata-x-either` re-exports big (most useful?) chunk of  `either`'s functions..

src/Mafia/Process.hs
jacobstanley: Hmm, but it will make defining equality downstream a problem as well.

I think I might change it to `IOException` instead

src/Mafia/Process.hs
jacobstanley: Yeah you're right, this should be fixed

domdere: It should have been `False` anyway, not `True`

mafia.cabal
jacobstanley: @EduardSergeev so it does! i'll just remove this altogether

src/Mafia/Process.hs
jacobstanley: The dangers of writing your own typeclass instances, humans are terrible

nhibberd:
- `mafia build` -- works
- Add a submodule in lib
  - `mafia build` -- `Adds to sandbox and mafia cache`
- Add to <project>.submodule file
- `mafia build` -- fails cant find dependency (`./cabal build` works)

nhibberd:
_project_.submodules file

> mafia watch test/test.hs -- --height 20
> mafia: _project_/lib/_dependency_: getDirectoryContents: does not exist (No such file or directory)

Should check if directory exists

jacobstanley:
I pushed a few more fixes, might fix these ^

src/Mafia/Process.hs
```diff
@@ -0,0 +1,329 @@
...
+forkWait :: IO a -> IO (IO (Either IOException a))
+forkWait io = do
+  mv <- newEmptyMVar
+
+  _  <- mask $ \restore ->
+    forkIO $ do
+      x <- try (restore io)
+      putMVar mv x
+
+  return (takeMVar mv)
```
tranma: hm, is this the same as the `wait`s in `async`, except `async` uses STM?

jacobstanley: yeah it's probably the same as `async` + `waitCatch` except that we're only catching `IOException`

This type of code is pretty easy to get wrong, so, happy to change it to use `async` instead. Although, that means I have the infuriating problem where we don't have an instance for `Eq SomeException`. Still, worth it I think.

tranma: perhaps you can use `tryIO` in http://hackage.haskell.org/package/errors-2.0.0/docs/Control-Error-Util.html to catch only IO exceptions?

jacobstanley: It's ok, it's not so bad to catch everything in this case.

Errors just weren't meant to be Eq'd I guess

markhibberd:
:hammer:

PR #5, opened 2015-09-30
First cut of mafia bootstrapping script

jacobstanley:
/cc @charleso (aka the shell scripting master)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "lib/p"]
 	path = lib/p
-	url = git@github.com:ambiata/p
+	url = https://github.com/ambiata/p
 [submodule "lib/disorder"]
 	path = lib/disorder
-	url = git@github.com:ambiata/disorder.hs
+	url = https://github.com/ambiata/disorder.hs
 [submodule "lib/x"]
 	path = lib/x
-	url = git@github.com:ambiata/x
+	url = https://github.com/ambiata/x
--- a/script/mafia
+++ b/script/mafia
@@ -1,0 +1,70 @@
+#!/bin/sh -eu
+
+fetch_latest () {
+  if [ -z ${MAFIA_TEST_MODE+x} ]; then
+    TZ=$(date +"%T")
+    curl --silent "https://raw.githubusercontent.com/ambiata/mafia/master/script/mafia?$TZ"
+  else
+    cat ../script/mafia
+  fi
+}
+
+run_upgrade () {
+  echo "Checking for a new version of mafia ..."
+  fetch_latest > /tmp/mafia
+
+  COMMIT_VERSION=$(git ls-remote https://github.com/ambiata/mafia | grep refs/heads/master | cut -f 1)
+  echo "# Version: $COMMIT_VERSION" >> /tmp/mafia
+
+  if ! cmp ./mafia /tmp/mafia >/dev/null 2>&1; then
+    echo "New version found and upgraded. You can now commit it to your git repo."
+    mv /tmp/mafia ./mafia
+    chmod +x ./mafia
+  else
+    echo "You have latest mafia script"
+  fi
+}
+
+exec_mafia () {
+  MAFIA_VERSION=$(awk '/^# Version: / { print $3; exit 0; }' $0)
+  MAFIA_PATH=$HOME/.ambiata/mafia/bin/mafia-$MAFIA_VERSION
+
+  [ -f "$MAFIA_PATH" ] || {
+    # Create a temporary directory which will be deleted when the script
+    # terminates. Unfortunately `mktemp` doesn't behave the same on
+    # Linux and OS/X so we need to try two different approaches.
+    MAFIA_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'exec_mafia')
+    trap "rm -rf \"$MAFIA_TEMP\"" EXIT
+
+    echo "Building mafia-$MAFIA_VERSION in $MAFIA_TEMP"
+
+    ( cd "$MAFIA_TEMP"
+
+      git clone https://github.com/ambiata/mafia
+      cd mafia
+
+      git reset --hard $MAFIA_VERSION
+
+      bin/bootstrap ) || exit $?
+
+    mkdir -p $(dirname $MAFIA_PATH)
+    cp "$MAFIA_TEMP/mafia/.cabal-sandbox/bin/mafia" "$MAFIA_PATH"
+  }
+
+  exec $MAFIA_PATH $@
+}
+
+#
+# The actual start of the script.....
+#
+
+if [ $# -gt 0 ]; then
+  MODE="$1"
+else
+  MODE=""
+fi
+
+case "$MODE" in
+upgrade) shift; run_$MODE "$@" ;;
+*) exec_mafia "$@"
+esac
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -1,0 +1,12 @@
+#!/bin/sh -eu
+
+git submodule update --init
+
+cabal sandbox init
+
+CABAL_SOURCES=$(find lib -maxdepth 4 ! -path lib/\*/bin/\* ! -path lib/\*/lib/\* -name \*.cabal | xargs -L 1 dirname)
+for CABAL_SOURCE in $CABAL_SOURCES; do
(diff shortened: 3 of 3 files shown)


amosr:
it is beautiful. what does it do other than update itself?

jacobstanley:
`mafia_exec` clones and builds the relevant version of mafia from github, and stores it in a well known location for later

tranma:
`mafia_exec`, awesome :sunglasses:

jacobstanley:
lol or `exec_mafia` even

tranma:
:gun:

.gitmodules
```diff
@@ -1,9 +1,9 @@
 [submodule "lib/p"]
 	path = lib/p
-	url = git@github.com:ambiata/p
+	url = https://github.com/ambiata/p
 [submodule "lib/disorder"]
 	path = lib/disorder
-	url = git@github.com:ambiata/disorder.hs
+	url = https://github.com/ambiata/disorder.hs
```
charleso, line 6: Was this intentional?

jacobstanley, line 6: Yeah, if someone was trying to build a mafia project, they shouldn't necessarily require github ssh credentials to be set up.

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+      done
+
+      cabal sandbox init
+
+      CABAL_SOURCES=$(find lib -maxdepth 4 ! -path lib/\*/bin/\* ! -path lib/\*/lib/\* -name \*.cabal | xargs -L 1 dirname)
+      for CABAL_SOURCE in $CABAL_SOURCES; do
+        cabal sandbox add-source -- $CABAL_SOURCE
+      done
+
+      cabal install )
```
charleso: I think you'll need `cabal install ) || exit $?` here

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+    mv /tmp/mafia ./mafia
+    chmod +x ./mafia
+  else
+    echo "You have latest mafia script"
+  fi
+}
+
+exec_mafia () {
+  MAFIA_VERSION=$(awk '/^# Version: / { print $3; exit 0; }' $0)
+  MAFIA_BASE=$HOME/.ambiata/mafia/bin/mafia
```
charleso: Bikeshedding, but `~/.mafia` or `~/.ambiata/mafia`?

nhibberd: Already using `~/.ambiata` for app-shop and box afaik

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+  MAFIA_PATH=$MAFIA_BASE-$MAFIA_VERSION
+
+  [ -f "$MAFIA_PATH" ] || {
+    MAFIA_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'exec_mafia')
+    trap "rm -rf \"$MAFIA_TEMP\"" EXIT
+
+    ( cd "$MAFIA_TEMP"
+
+      git clone https://github.com/ambiata/mafia
```
markhibberd, line 43: Not sure if it is worth it, but only needs to a sparse checkout, might speed things up. @charleso can probably wow you with git.

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+      cd mafia
+
+      git reset --hard $MAFIA_VERSION
+
+      SUBMODULES=$(find lib -mindepth 1 -maxdepth 1)
+      for SUBMODULE in $SUBMODULES; do
+        git submodule update --init -- $SUBMODULE
```
markhibberd: Any reason to not just do 1 call to do them all?

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+    trap "rm -rf \"$MAFIA_TEMP\"" EXIT
+
+    ( cd "$MAFIA_TEMP"
+
+      git clone https://github.com/ambiata/mafia
+      cd mafia
+
+      git reset --hard $MAFIA_VERSION
+
+      SUBMODULES=$(find lib -mindepth 1 -maxdepth 1)
```
charleso: For all the building of mafia should it rely on a script _in_ the repository?

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+      CABAL_SOURCES=$(find lib -maxdepth 4 ! -path lib/\*/bin/\* ! -path lib/\*/lib/\* -name \*.cabal | xargs -L 1 dirname)
+      for CABAL_SOURCE in $CABAL_SOURCES; do
+        cabal sandbox add-source -- $CABAL_SOURCE
+      done
+
+      cabal install )
+
+    mkdir -p $(dirname $MAFIA_PATH)
+    cp "$MAFIA_TEMP/mafia/.cabal-sandbox/bin/mafia" "$MAFIA_PATH"
+  }
```
markhibberd, line 52: Not sure, but I think a lot of this might be easier to maintain outside of this script, as in, if this script just did a checkout, and than ran ./bootstrap or similar - would make it easier to change etc.... Not a big deal though.

script/mafia
jacobstanley, line 43: Hmm, so I tried `git clone --depth 1` but it won't necessarily have the commit hash we need

script/mafia
jacobstanley: would that be `git submodule update --init` ?

`git submodule update --init --recursive` goes too far as far as I can tell

script/mafia
jacobstanley: I tried this out locally and `set -e` seems to be enough (I don't need to exit thing)

script/mafia
markhibberd: I think this should have the version in it, so you can run multiple versions for different projects at same time if necessary.

jacobstanley: Yeah I figured we'll end up a huge number of `~/.foo` directories if we don't put all the the ambiata stuff together?

script/mafia
markhibberd: It won't be in all shells, when it comes to `set -e` and subshells, the only rule is there isn't any rules.

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+  fi
+}
+
+exec_mafia () {
+  MAFIA_VERSION=$(awk '/^# Version: / { print $3; exit 0; }' $0)
+  MAFIA_BASE=$HOME/.ambiata/mafia/bin/mafia
+  MAFIA_PATH=$MAFIA_BASE-$MAFIA_VERSION
+
+  [ -f "$MAFIA_PATH" ] || {
+    MAFIA_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'exec_mafia')
```
olorin, line 36: The second `mktemp` breaks on GNU systems; I assume it's correct for some other variant? Comment might be handy.

markhibberd:
Couple of comments in line - not concerned with the shallow clone thing, was just an idea - :+1:

script/mafia
markhibberd: Yeh, the first one.

script/mafia
charleso: > I think this should have the version in it, so you can run multiple versions for different projects at same time if necessary.

I might be tempted to do a bare, full clone to a single place, and then `git clone --no-checkout -s mafia mafia-$version && cd mafia-$version && git reset --hard $version` or something like that

script/mafia
```diff
@@ -0,0 +1,75 @@
...
+    trap "rm -rf \"$MAFIA_TEMP\"" EXIT
```
olorin, line 37: Maybe SIG{INT,TERM} as well? Is usually idiomatic to clean up after the more polite signals.

jacobstanley, line 37: I read somewhere that EXIT subsumes this?

charleso:
:+1:

script/mafia
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  exec $MAFIA_PATH $@
+}
+
+#
+# The actual start of the script.....
+#
+
+[ $# -gt 0 ] || fail
```
charleso: Is `fail` a thing?

olorin:
:shaved_ice:

script/mafia
damncabbage, line 37: Confirmed; I just tested that at least `/bin/sh` runs the command on SIGHUP, -INT and -TERM.

PR #10, opened 2015-10-12
.submodules file contains paths which are relative to itself (not the .git root)

jacobstanley:


markhibberd:
Mmmm, maybe I am mis-reading this, but it looks like the old way is correct?

The submodule file should contain file paths relative to git root.

nhibberd:
There were old versions of cabal-extras that behaved relative to the project name. :older_man:

jacobstanley:
Makes sense, ok, killing this

PR #11, opened 2015-10-12
Unregister packages when a cabal file is no longer referenced

jacobstanley:


main/mafia.hs
```diff
@@ -327,20 +328,65 @@ mkFileMap = Map.fromList . fmap (\path -> (takeFileName path, path))
...
-     -> do liftIO (T.hPutStrLn stderr ("Cache: Removed " <> takeFileName file))
+    Delete file -> do
+      liftIO (T.hPutStrLn stderr ("Cache: Removed " <> takeFileName file))
 
-    Update src _
-     -> do rel <- fromMaybe src <$> makeRelativeToCurrentDirectory src
-           liftIO (T.hPutStrLn stderr ("Cache: Modified " <> rel))
+    Update src _ -> do
+      rel <- fromMaybe src <$> makeRelativeToCurrentDirectory src
+      liftIO (T.hPutStrLn stderr ("Cache: Modified " <> rel))
```
jacobstanley, line 336: ^ sorry this is just a style fixup so all the case expressions are the same in here

tranma:
does this unregister dependencies when they are removed from the .cabal file?

markhibberd:
:8ball:

jacobstanley:
@tranma no it's only for submodule dependencies

jacobstanley:
Hmm I think this might have a few broken corner cases, it's one of the reasons I wrote the test in the other PR. I'm going to expand that test to include messing with the `<project>.submodules` file and then see if this still works.

jacobstanley:
I'm going to put up a different PR for this which actually works, and has more thorough tests.

PR #12, opened 2015-10-13
Fixed bug where cabal would trash its sandbox and be unable to add/delete-source

jacobstanley:
If you ever delete a directory which you have previously added to a cabal sandbox via add-source then cabal will be unable to add-source or remove-source from then on.

The fix is a rather complex workaround unfortunately :( We create the directory which cabal is looking for and put a dummy cabal file in it.

### DIFF
--- a/src/Mafia/Cabal.hs
+++ b/src/Mafia/Cabal.hs
@@ -1,0 +1,78 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+module Mafia.Cabal
+  ( SandboxDir
+  , CabalError(..)
+  , repairIndexFile
+  ) where
+
+import           Control.Exception (IOException)
+import           Control.Monad.Catch (catch)
+import           Control.Monad.IO.Class (MonadIO(..))
+
+import           Data.ByteString (ByteString)
+import qualified Data.ByteString.Lazy as L
+import qualified Data.Text.Encoding as T
+
+import qualified Codec.Archive.Tar as Tar
+
+import           Mafia.IO
+import           Mafia.Path
+
+import           P
+
+import           System.IO (IO)
+
+import           X.Control.Monad.Trans.Either
+
+------------------------------------------------------------------------
+
+type SandboxDir = Directory
+
+data CabalError =
+    IndexFileNotFound File
+  | CorruptIndexFile Tar.FormatError
+  deriving (Show)
+
+------------------------------------------------------------------------
+
+readIndexFile :: MonadIO m => File -> EitherT CabalError m ByteString
+readIndexFile file = do
+  mbs <- readBytes file
+  hoistEither $ maybe (Left (IndexFileNotFound file)) Right mbs
+
+readEntries :: Monad m => ByteString -> EitherT CabalError m [Tar.Entry]
+readEntries bs = do
+  let tar = Tar.read (L.fromStrict bs)
+  hoistEither $ Tar.foldEntries (\x -> fmap (x :)) (Right []) (Left . CorruptIndexFile) tar
+
+isTreeRef :: Tar.Entry -> Bool
+isTreeRef x = Tar.entryPath x == "local-build-tree-reference/"
+
+fromEntry :: Tar.Entry -> Maybe Directory
+fromEntry x =
+  case Tar.entryContent x of
+    Tar.OtherEntryType _ lbs _ | isTreeRef x -> Just (T.decodeUtf32LE (L.toStrict lbs))
+    _                                        -> Nothing
+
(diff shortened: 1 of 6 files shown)


markhibberd:
:100: for test, pretty ugly in some spots, but the cost of cabal'ing I guess. Seems like something worthy of pinging Duncan with to demonstrate the extent we work around things.

jacobstanley:
Any thoughts on improving the ugly?

jacobstanley:
Ack regarding Duncan

markhibberd:
Not sure it would result in less pain, but can you just implement delete source yourself, strip out of timestamps file and config?

jacobstanley:
Unfortunately it stores stuff in the `00-index.tar` in what appears to be some binary format (or maybe just UTF16?)

I had a look at the cabal source code trying to follow what's going on and it doesn't seem obvious.

The timestamps file doesn't really seem to be involved in the add-source/delete-source operation. I tried messing around with it and it doesn't have any effect.

jacobstanley:
This is involved somewhere along the line:

``` haskell
-- | Convert a 'FilePath' to a lazy 'ByteString'. Each 'Char' is
-- encoded as a little-endian 'Word32'.
filePathToByteString :: FilePath -> BS.ByteString
filePathToByteString p =
  BS.pack $ foldr conv [] codepts
  where
    codepts :: [Word32]
    codepts = map (fromIntegral . ord) p

    conv :: Word32 -> [Word8] -> [Word8]
    conv w32 rest = b0:b1:b2:b3:rest
      where
        b0 = fromIntegral $ w32
        b1 = fromIntegral $ w32 `shiftR` 8
        b2 = fromIntegral $ w32 `shiftR` 16
        b3 = fromIntegral $ w32 `shiftR` 24
```

olorin:
Yeah, it's UTF-32.

```
âˆƒ hexdump -C local-build-tree-reference 
00000000  2f 00 00 00 68 00 00 00  6f 00 00 00 6d 00 00 00  |/...h...o...m...|
00000010  65 00 00 00 2f 00 00 00  73 00 00 00 69 00 00 00  |e.../...s...i...|
00000020  6f 00 00 00 2f 00 00 00  73 00 00 00 72 00 00 00  |o.../...s...r...|
00000030  63 00 00 00 2f 00 00 00  74 00 00 00 61 00 00 00  |c.../...t...a...|
00000040  74 00 00 00 6f 00 00 00  6f 00 00 00 69 00 00 00  |t...o...o...i...|
00000050  6e 00 00 00 65 00 00 00  2f 00 00 00 6c 00 00 00  |n...e.../...l...|
00000060  69 00 00 00 62 00 00 00  2f 00 00 00 62 00 00 00  |i...b.../...b...|
00000070  6f 00 00 00 6e 00 00 00  65 00 00 00 73 00 00 00  |o...n...e...s...|
00000080
```

jacobstanley:
Hmm, maybe it's not infeasible to do it by hand

markhibberd:
Fair enough, agreed it sounds like too much. :+1: on current approach.

jacobstanley:
For future reference, the sandbox index code is here: https://github.com/haskell/cabal/blob/master/cabal-install/Distribution/Client/Sandbox/Index.hs

jacobstanley:
I had a crack at repairing the `00-index.tar` file and it seems to work, is this a better solution?

jacobstanley:
Bump, cabal salvation is at hand

src/Mafia/Cabal.hs
```diff
@@ -0,0 +1,78 @@
...
+    Tar.OtherEntryType _ lbs _ | isTreeRef x -> Just (T.decodeUtf32LE (L.toStrict lbs))
+    _                                        -> Nothing
+
+------------------------------------------------------------------------
+
+repairIndexFile :: SandboxDir -> EitherT CabalError IO ()
+repairIndexFile = filterEntries doesDirectoryExist
+
+filterEntries :: (Directory -> IO Bool) -> File -> EitherT CabalError IO ()
+filterEntries pred sandboxDir = do
```
charleso, line 65: `pred` not used?

charleso, line 65: > pred not used?

Ignore me

src/Mafia/Cabal.hs
```diff
@@ -0,0 +1,78 @@
...
+  let indexFile  = sandboxDir </> "packages/00-index.tar"
```
charleso, line 66: As discussed not 100% whether it makes sense to fail hard if this file doesn't exist or not. On the one hand it means we can spot issues quicker, but I'm vaguely thinking of cabal's track record of breaking compatibility and what that would mean for `mafia` projects one year from now. I'm hoping will be a different beast by then anyway, so just ignore me. :)

charleso:
:100: Let's see how far the :rabbit: hole goes.

PR #13, opened 2015-10-16
Unregister packages when a .cabal file is no longer referenced

jacobstanley:


markhibberd:
:100: Epic test.

domdere:
If it could also generate code in those libraries we could setup up basket to catch all the useful libraries CI might generate

charleso:
This is phenomenal :100: \* :100:

PR #14, opened 2015-10-16
Use `git submodule` to detect which lib/ folders to use

jacobstanley:
Not sure if this is controversial or not?

This changes mafia to use _only_ the registered submodules from the `lib/` directory when building.

This should fix the situation where you switch branches and a submodule is unregistered in the process.

### DIFF
--- a/mafia.cabal
+++ b/mafia.cabal
@@ -40,6 +40,7 @@ library
                     BuildInfo_ambiata_mafia
                     Mafia
                     Mafia.Cabal
+                    Mafia.Git
                     Mafia.IO
                     Mafia.Path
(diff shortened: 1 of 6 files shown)


charleso:
As discussed IRL it's tricky to do this any other way without being explicit.

The biggest downside I see even for us is that if you're trying something new locally you'll _have_ to use `git submodule add` instead of, say, `git clone ... lib/...`, but I can't really see that being a huge problem.

I guess I wonder whether at some point we should detect if git is being used, and fallback to other safe behaviour. (It might do that now, in which case ignore me).

Still, :+1: for our current use-case.

PR #18, opened 2015-10-27
Added -a/--all to `mafia quick` and `mafia watch`

jacobstanley:
/cc @charleso @nhibberd

This adds support for loading the source of all libraries when using quick/watch. Seems like `ghci` is smart enough to only load the things it needs, pretty cool.

Not sure if `src`, `gen` and `dist/build/autogen` are the right sub-directories to be including or not? Seems to work ok :)

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -35,7 +35,7 @@ import           System.IO (IO, stdout, stderr, putStrLn, print)
 import           X.Control.Monad.Trans.Either (EitherT(..), hoistEither, firstEitherT)
 import           X.Options.Applicative (Parser, CommandFields, Mod)
 import           X.Options.Applicative (SafeCommand(..), RunType(..))
-import           X.Options.Applicative (argument, textRead, metavar, help, long, short, option)
+import           X.Options.Applicative (argument, textRead, metavar, help, long, short, option, flag')
 import           X.Options.Applicative (dispatch, orDie, subparser, safeCommand, command')
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


nhibberd:
woah

main/mafia.hs
```diff
@@ -257,6 +274,19 @@ ghciArgs extraIncludes path = do
             <> (fmap ("-package-db=" <>) databases)
             <> [ path ]
 
+reifyInclude :: GhciInclude -> EitherT MafiaViolation IO [Directory]
+reifyInclude = \case
+  Directory dir -> return [dir]
+  AllLibraries  -> do
+    absDirs <- Set.toList <$> getSubmoduleSources
+    relDirs <- mapM tryMakeRelativeToCurrent absDirs
+    return [ dir </> sub | dir <- relDirs
```
nhibberd, line 283: Not sure about this bit

charleso:
Not sure what the downside/implications are, but happy to be a guinea :pig: to try this out :+1:

main/mafia.hs
jacobstanley, line 283: Would this be better?

``` haskell
    return $ do
      dir <- relDirs
      sub <- ["src", "test", "gen", "dist/build/autogen"]
      return (dir </> sub)
```

jacobstanley, line 283: Or maybe this?

``` haskell
    return $ (</>) <$> relDirs <*> ["src", "test", "gen", "dist/build/autogen"]
```

tranma, line 283: :fearful:

jacobstanley, line 283: Ok, talked to @nhibberd IRL, sticking with the list comprehension

PR #19, opened 2015-11-04
Always read/write files using UTF-8 encoding

jacobstanley:
This is an attempt to fix the farm build issue: https://github.com/ambiata/eminence/pull/183

/cc @tmcgilchrist

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -448,13 +448,13 @@ tryUnregisterPackage cabalFile = do
 
 invalidateCache :: MonadIO m => File -> m ()
 invalidateCache cabalFile = do
-  mtxt <- readText cabalFile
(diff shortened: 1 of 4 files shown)


olorin:
:100: for always writing stuff in UTF-8, but why not use `Data.Text.IO.{write,read}File`?

jacobstanley:
@olorin as far as I can tell `{write,read}File` uses whatever the system encoding is (i.e. the `LANG` environment variable)

olorin:
Yep, you're correct, I was misreading the docs. :shaved_ice:

tmcgilchrist:
It's failing for other reasons now, so I suppose that's classified as success.

jacobstanley:
Added a yak for this: https://github.com/ambiata/yak/issues/75

PR #21, opened 2015-11-11
Adding mafia bench

nhibberd:
@jystic @markhibberd 
not sure what / where I should add tests etc


jacobstanley:
@charleso would be proud :green_book:

PR #23, opened 2015-12-08
Migrate to 7.10

thumphries:
Build now succeeds with 7.8.4 and 7.10.2.

The only clash was in `Setup.hs`, worked around with `time-locale-compat`, and the changes in ambiata/p#55.

cc ambiata/yak#63

PR #28, opened 2015-12-20
Keep a ghc version specific sandbox

jacobstanley:
This updates mafia to create a sandbox specific to the version of ghc it finds on the path, so you might end up with the following sort of structure:

```
.cabal-sandbox/7.8.4
.cabal-sandbox/7.10.2
```

One annoying thing is that `cabal.sandbox.config` must be updated to point to the correct sandbox. We could work around this by setting the `CABAL_SANDBOX_CONFIG` environment variable when using mafia, but it would mean that if people use cabal directly then it wouldn't work.

I was also able to build mafia on 7.10 by just widening the bounds on time and upgrading p, so I have done that too.

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -39,7 +39,7 @@ import           System.IO (IO, stdout, stderr, putStrLn, print)
 import           System.IO.Temp (withTempDirectory)
 
 import           X.Control.Monad.Trans.Either (EitherT, pattern EitherT, runEitherT)
-import           X.Control.Monad.Trans.Either (hoistEither, firstEitherT)
+import           X.Control.Monad.Trans.Either (hoistEither, firstEitherT, left)
 import           X.Control.Monad.Trans.Either.Exit (orDie)
 import           X.Options.Applicative (Parser, CommandFields, Mod)
 import           X.Options.Applicative (SafeCommand(..), RunType(..))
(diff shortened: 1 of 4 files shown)


main/mafia.hs
```diff
@@ -618,8 +657,8 @@ cabal_ cmd args = do
 
 sandbox :: ProcessResult a => Argument -> [Argument] -> EitherT MafiaViolation IO a
 sandbox cmd args = do
-  sandboxDir <- getSandboxDir
-  cabal "sandbox" $ ["--sandbox", sandboxDir] <> (cmd:args)
+  _ <- initSandbox
+  cabal "sandbox" (cmd:args)
```
jacobstanley, line 661: It turns out that passing `--sandbox <dir>` to cabal has no effect unless you're initialising a sandbox, so I have removed it. The only configurable way of changing the sandbox to use is to set the `CABAL_SANDBOX_CONFIG` environment variable.

jacobstanley:
This now includes everything in https://github.com/ambiata/mafia/pull/23 as well.

charleso:
As discussed, seems legit. Not sure about the scraping of the cabal file, possibly symlinks or something similar might be better, but a good start. :+1:

domdere:
:+1:

PR #29, opened 2015-12-21
Install ghcid automagically

charleso:
/cc @jystic 

I'm assuming the latest version `0.5` is OK


jacobstanley:
Wow this is awesome, `installBinary` is pretty neat :key:

PR #30, opened 2015-12-22
Extract small modules where possible

charleso:
/cc @jystic 

This took way longer than I had hoped. I went down the rabbit hole of trying to tease apart the errors types, but in the end this is pretty much as it was.

Definitely open to suggestion on module names and splits.

### DIFF
--- a/src/Mafia/Home.hs
+++ b/src/Mafia/Home.hs
@@ -1,0 +1,59 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
+module Mafia.Home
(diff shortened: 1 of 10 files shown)


src/Mafia/Project.hs
```diff
@@ -0,0 +1,41 @@
+{-# LANGUAGE NoImplicitPrelude #-}
```
charleso, line 1: Not sure if this really deserves a separate module. Is there a better place for it?

src/Mafia/Home.hs
```diff
@@ -0,0 +1,59 @@
+{-# LANGUAGE LambdaCase #-}
```
charleso, line 1: I hate this name, wasn't sure what to call it

jacobstanley:
:100: This is a great first step, I'm sure we can tease the errors apart over time.

PR #31, opened 2015-12-22
Include any remaining cached hoogle packages in search

charleso:
/cc @jystic


jacobstanley:
Looks great :racehorse:

PR #32, opened 2015-12-22
Crude work-around for when (savages) clean.

markhibberd:
Currently if you run cabal clean the configure
doesn't not get triggered on subsequent commands
because the cabal file isn't detemined to be out
of date. This just checks if dist exists and
forcibly configures in that case, this is crude
and there are probably holes in it, but covers
a common case.


markhibberd:
Warning not actually (sandboxed), so build-success may be just luck.

jacobstanley:
If this works for you locally then I'm happy :+1:

PR #33, opened 2015-12-22
More module reorganisation

jacobstanley:
A little bit more reorg before I start to try and integrate my sandboxing upgrades


jacobstanley:
There should be no logic changes here, just moving things around.

jacobstanley:
@markhibberd do you think you can take a look at this? it should fairly straightforward just be moving things around.

markhibberd:
:yen:

charleso:
Nice :8ball:

PR #34, opened 2016-01-03
Nix-style package management

jacobstanley:
Adds support for nix-style package management by creating a sandbox in a global location for each package to be installed, this means they can be shared across sandboxes using symlinks.

The idea comes from https://github.com/hvr/hackage-matrix-builder/ which does something similar.

Not sure it's ready for wider consumption just yet but I want to see if it will build on CI.

### DIFF
--- a/src/Mafia/Cabal/Dependencies.hs
+++ b/src/Mafia/Cabal/Dependencies.hs
@@ -21,6 +21,7 @@ import           Mafia.Cabal.Index
 import           Mafia.Cabal.Package
 import           Mafia.Cabal.Process (cabalFrom)
 import           Mafia.Cabal.Types
+import           Mafia.Cabal.Version
 import           Mafia.IO
 import           Mafia.Package
 import           Mafia.Path
@@ -86,6 +87,8 @@ reifyPackageRefs refs =
 
 findRevDeps :: [Directory] -> EitherT CabalError IO [RevDeps]
 findRevDeps sdirs = do
+  checkCabalVersion
+
   EitherT . withSystemTempDirectory "mafia-deps-" $ \tmp0 -> runEitherT $ do
     dir <- getCurrentDirectory
 
--- a/src/Mafia/Cabal/Dependencies.hs
+++ b/src/Mafia/Cabal/Dependencies.hs
@@ -38,7 +38,7 @@ import           X.Control.Monad.Trans.Either
 
 findDependencies :: [SourcePackage] -> EitherT CabalError IO [Package]
 findDependencies spkgs = do
-  fromRevDeps spkgs <$> findRevDeps (fmap spDirectory spkgs)
+  fromRevDeps spkgs <$> findRevDeps spkgs
 
 fromRevDeps :: [SourcePackage] -> [RevDeps] -> [Package]
 fromRevDeps spkgs rdeps =
@@ -85,8 +85,17 @@ reifyPackageRefs refs =
         mkPackage ref (mapMaybe (\d -> Map.lookup d pkgs) deps)
   in pkgs
 
-findRevDeps :: [Directory] -> EitherT CabalError IO [RevDeps]
-findRevDeps sdirs = do
+mapFromList :: Ord k => (v -> k) -> [v] -> Map k v
+mapFromList f xs = Map.fromList (List.zip (fmap f xs) xs)
+
+toGraphKey :: RevDeps -> (RevDeps, PackageId, [PackageId])
+toGraphKey rev = (rev, refId (revRef rev), revDeps rev)
+
+fromGraphKey :: (RevDeps, PackageId, [PackageId]) -> PackageRef
+fromGraphKey (rev, _, _) = revRef rev
+
+findRevDeps :: [SourcePackage] -> EitherT CabalError IO [RevDeps]
+findRevDeps spkgs = do
   checkCabalVersion
 
   EitherT . withSystemTempDirectory "mafia-deps-" $ \tmp0 -> runEitherT $ do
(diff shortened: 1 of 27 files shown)


charleso:
Have started testing this locally and so far so good. This is game changing!

src/Mafia/Init.hs
```diff
@@ -2,193 +2,163 @@
...
+  packageDB     <- firstEitherT InitCabalError getPackageDB
+  packages      <- getDirectoryListing (RecursiveDepth 0) packageDB
+  packagesExist <- and <$> traverse doesFileExist packages
+  case packagesExist of
+    True  -> return PackagesOk
+    False -> return PackagesBroken
+
+------------------------------------------------------------------------
+
+data MafiaState =
```
charleso, line 144: Worth having a separate module for this?

src/Mafia/Install.hs
```diff
@@ -0,0 +1,214 @@
...
+  globals   <- installGlobalDependencies spkgs
+  _         <- firstEitherT InstallCabalError initSandbox
+  packageDB <- firstEitherT InstallCabalError getPackageDB
+
+  ignoreIO $ removeDirectoryRecursive packageDB
+  createDirectoryIfMissing True packageDB
+
+  env  <- getPackageEnv
+  mapM_ (link packageDB env) globals
+  Hush <- firstEitherT InstallCabalError $ cabal "sandbox" ["hc-pkg", "recache"]
```
charleso, line 87: Worth adding a comment about what this does?

src/Mafia/Package.hs
```diff
@@ -48,15 +57,38 @@ packageIdTuple (PackageId n v) =
...
     component = do
       cs <- Parse.munch1 Char.isAlphaNum
       if all Char.isDigit cs then Parse.pfail else return cs
+
+parseVersion :: Text -> Maybe Version
+parseVersion =
+  parseLongestMatch Version.parseVersion
+
+parseLongestMatch :: Parse.ReadP a -> Text -> Maybe a
+parseLongestMatch p =
```
charleso, line 78: Yeah I was surprised this didn't live somewhere already?

charleso:
Might be good to add a doc file about how the shared sandboxes work?

src/Mafia/Install.hs
```diff
@@ -0,0 +1,214 @@
...
+link :: Directory -> PackageEnv -> Package -> EitherT InstallError IO ()
+link db env p@(Package (PackageRef pid _ _) _ _) = do
+  let pcfg = packageConfigPath env p
+  unlessM (doesFileExist pcfg) (install env p)
+
+  let dest = db </> renderPackageId pid <> ".conf"
+  liftIO $ createSymbolicLink (T.unpack pcfg) (T.unpack dest)
+
+install :: PackageEnv -> Package -> EitherT InstallError IO ()
+install env p@(Package (PackageRef pid _ msrc) deps _) = do
```
charleso: Some high level doc on this function might be good too

src/Mafia/Cabal/Dependencies.hs
```diff
@@ -0,0 +1,191 @@
...
+
+parseRevDeps :: Text -> Either CabalError [RevDeps]
+parseRevDeps =
+  first (CabalParseError . T.pack) .
+  traverse parseDep .
+  List.drop 1 .
+  List.dropWhile (/= "In order, the following would be installed:") .
+  T.lines
+
+parseDep :: Text -> Either String RevDeps
```
charleso: This is probably worth some unit-level tests

charleso:
More tests, but :100:

src/Mafia/Cabal/Version.hs
```diff
@@ -0,0 +1,59 @@
...
+        Just ver -> return ver
+
+parseCabalVersion :: Text -> Maybe Version
+parseCabalVersion out =
+  case T.words out of
+    ("cabal-install" : "version" : version : _) -> parseVersion version
+    _                                           -> Nothing
+
+-- TODO not sure if maybe we should have no upper bounds here?
+-- TODO maybe mafia should just install its own cabal?
```
charleso, line 43: Yeah I guess as discussed that would be the next thing. And then mafia can...

conquer the world

src/Mafia/IO.hs
```diff
@@ -161,8 +172,29 @@ findExecutable name = liftIO $ do
...
 getHomeDirectory :: MonadIO m => m Directory
 getHomeDirectory = T.pack `liftM` liftIO Directory.getHomeDirectory
+
+------------------------------------------------------------------------
+-- Concurrency
+
+mapConcurrentlyE :: Traversable t => (a -> EitherT x IO b) -> t a -> EitherT x IO (t b)
+mapConcurrentlyE io xs = do
+  ys <- liftIO (Async.mapConcurrently (runEitherT . io) xs)
+  hoistEither (sequence ys)
```
tranma, line 200: nice

src/Mafia/Package.hs
jacobstanley, line 78: Yeah I'm tempted to rewrite the lot as an attoparsec parser just for consistency

src/Mafia/Init.hs
jacobstanley, line 144: Yeah probably, originally it was the only thing in here and was more like `InitState`

jacobstanley, line 144: Hmm you know to be fair this is the only thing the init module really does (keep track of the dependency state). Maybe this should just be called something different?

thumphries:
This looks awesome!

src/Mafia/Cabal/Version.hs
markhibberd, line 43: I think installing own cabal is tremendously dangerous/not a good idea. As it stands using cabal and mafia at the same time is standard faire, and risking inconsistent versions could have particularly dire consequences. At least in the naive version of just installing its own and using it regardless of what is on path.

Issue #36, opened 2016-01-04
Does not support multiple mafias initialising at same time


nhibberd:

olorin:
My default solution would just be have mafia wait on a lockfile under ~/.ambiata; would that solve the current issue?

jacobstanley:
This is where the problem is:

``` sh
      mkdir -p $(dirname $MAFIA_PATH)
      cp "$MAFIA_TEMP/mafia/.cabal-sandbox/bin/mafia" "${MAFIA_PATH}.tmp"
      mv "${MAFIA_PATH}.tmp" "$MAFIA_PATH"
```

The issue is that `$MAFIA_PATH` is not random, and it should be really.

jacobstanley:
well `$MAFIA_PATH` shouldn't be random, but `${MAFIA_PATH}.tmp` should be

PR #37, opened 2016-01-04
script/mafia is now properly atomic, this time for sure

jacobstanley:
This should in theory fix https://github.com/ambiata/mafia/issues/36

/cc @nhibberd

### DIFF
--- a/script/mafia
+++ b/script/mafia
@@ -32,7 +32,9 @@ exec_mafia () {
     # If we can't find the mafia version, then we need to upgrade the script.
     run_upgrade
   else
-    MAFIA_PATH=$HOME/.ambiata/mafia/bin/mafia-$MAFIA_VERSION
+    MAFIA_BIN=$HOME/.ambiata/mafia/bin
+    MAFIA_FILE=mafia-$MAFIA_VERSION
+    MAFIA_PATH=$MAFIA_BIN/$MAFIA_FILE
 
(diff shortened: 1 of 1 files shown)


olorin:
LGTM :+1:

nhibberd:
:fallen_leaf:

script/mafia
```diff
@@ -41,7 +43,11 @@ exec_mafia () {
       MAFIA_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'exec_mafia')
       trap "rm -rf \"$MAFIA_TEMP\"" EXIT
 
-      echo "Building mafia-$MAFIA_VERSION in $MAFIA_TEMP"
+      # Create a temporary file in MAFIA_BIN so we can do an atomic copy/move dance.
+      MAFIA_PATH_TEMP=$(mktemp --tmpdir=$MAFIA_BIN $MAFIA_FILE-XXXXXX 2>/dev/null || TMPDIR=$MAFIA_BIN mktemp -t $MAFIA_FILE)
+      trap "rm -f \"$MAFIA_PATH_TEMP\"" EXIT
```
charleso, line 48: I _think_ that this trap will override the one above. Would that be a problem?

jacobstanley, line 48: yeah that would be bad, it should probably be a single trap to a function call, will fix it up

PR #41, opened 2016-01-06
Ensure happy/alex can be added to the PATH

charleso:
/cc @jystic @thumphries


jacobstanley:
:nut_and_bolt: once you check the hoogle upgrade works

PR #43, opened 2016-01-06
Add support for parsing the reinstall section of a package install plan

jacobstanley:
I thought reinstalls could never happen because we always had a fresh
sandbox. It turns out they can because there are a few packages like
process, filepath, directory, etc that are in the global package
database which is installed by default.

Ideally we could just exclude the global package database alltogether
but we need it for base and possibly ghc.

/cc @charleso 

This needs to be merged before we can proceed with the upgrade on `x` because `x-conduit` does a reinstall of `process`.


charleso:
:+1:

PR #44, opened 2016-01-07
Use https for twine submodule

jacobstanley:
Without this mafia doesn't work on travis


olorin:
:+1:

nhibberd:
We generally use ssh for all our submodules, any reason these are https?

jacobstanley:
@nhibberd it has to be so that mafia can be built without having github authorisation, this is the only special case i think

nhibberd:
Ye touche nvm

PR #45, opened 2016-01-08
Fix annoyance where add-source dependencies are built twice

jacobstanley:
People might have noticed that sometimes on the build bot, source dependencies are being built twice.

This is because an older version of mafia has called 'cabal sandbox add-source' for each of the source dependencies and cabal thinks that it needs to build them, even though that have already been built.

You can fix this by just clearing the workspace on jenkins, but this fix provides a smoother upgrade and also prevents against problems with users calling add-source themselves.

/cc @charleso @nhibberd @markhibberd

### DIFF
--- a/src/Mafia/Cabal/Index.hs
+++ b/src/Mafia/Cabal/Index.hs
@@ -2,8 +2,9 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 module Mafia.Cabal.Index
-  ( repairIndexFile
-  , createIndexFile
+  ( createIndexFile
+  , readIndexFile
+  , repairIndexFile
   ) where
 
 import           Control.Exception (IOException)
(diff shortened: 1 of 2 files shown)


nhibberd:
Sounds legit :facepunch:

thumphries:
if a user has manually done a `sandbox add-source`, it'd be pretty surprising to have it clobbered silently. could we print a warning message too?

jacobstanley:
@thumphries yep i can do that

jacobstanley:
/cc @thumphries it now prints out something like:

```
Removing add-source dependency: lib/disorder/disorder-aeson
Removing add-source dependency: lib/disorder/disorder-core
Removing add-source dependency: lib/disorder/disorder-corpus
Removing add-source dependency: lib/disorder/disorder-fsm
Removing add-source dependency: lib/disorder/disorder-lens
Removing add-source dependency: lib/p
Removing add-source dependency: lib/twine
Removing add-source dependency: lib/x/x-aeson
Removing add-source dependency: lib/x/x-attoparsec
Removing add-source dependency: lib/x/x-conduit
Removing add-source dependency: lib/x/x-either
Removing add-source dependency: lib/x/x-eithert
Removing add-source dependency: lib/x/x-exception
Removing add-source dependency: lib/x/x-htoml
Removing add-source dependency: lib/x/x-optparse
Removing add-source dependency: lib/x/x-templatehaskell
```

thumphries:
sweet! thanks for that, was bound to get me eventually :shaved_ice:

PR #46, opened 2016-01-08
Install binaries with all their support files

thumphries:
Replaces `cabal install; cp` with `cabal install --prefix=$MAFIABIN/packageName-0.1.1`. All the supporting files (happy templates etc) get copied properly, and we return `$MAFIABIN/packageName-0.1.1/bin` instead.

Currently this will break if an old version of mafia has installed stuff, I'll add a workaround...

### DIFF
--- a/src/Mafia/Home.hs
+++ b/src/Mafia/Home.hs
@@ -21,10 +21,11 @@ import           Mafia.Process
 
 import           P
 
+import           System.FileLock (SharedExclusive (..), lockFile, unlockFile)
 import           System.IO (IO)
 import           System.IO.Temp (withTempDirectory)
 
-import           X.Control.Monad.Trans.Either (EitherT, pattern EitherT, runEitherT)
+import           X.Control.Monad.Trans.Either (EitherT, pattern EitherT, runEitherT, bracketEitherT')
 
 
 getMafiaHome :: MonadIO m => m Directory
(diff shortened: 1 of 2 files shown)


thumphries:
tested with happy, alex and hoogle

thumphries:
@charleso discussed this with @jystic briefly, but this will break existing installs of hoogle when old and new mafia versions are used together. old mafia will reach for `hoogle-version/hoogle` instead of `hoogle-version/bin/hoogle`. Could hack around this by symlinking, or just let it burn and tell people to upgrade mafia.

inclined to just break it

charleso:
> inclined to just break it

I'm ok with that.

src/Mafia/Home.hs
```diff
@@ -51,30 +51,29 @@ installBinary package deps = do
...
-        createDirectoryIfMissing True path
-        getDirectoryListing (RecursiveDepth 1) (T.pack sandboxDir </> ".cabal-sandbox" </> "bin") >>= \p ->
-          for_ p $ \f -> copyFile f (path </> takeFileName f)
-        return path
+  liftIO (doesFileExist prefix) >>= \b -> when b $ do
+    renameFile prefix (prefix <> ".tmp")
+    createDirectoryIfMissing True prefix
+    renameFile (prefix <> ".tmp") (prefix </> (unPackageName . pkgName) package)
+  exists <- liftIO (doesDirectoryExist path)
+  unless exists $ do
```
charleso: Don't need this `do` /supernit

charleso:
:+1: Thanks @thumphries 

Might be worth making a note about why we're keeping the sandbox, as it might not be immediately obvious unless you know.

src/Mafia/Home.hs
thumphries: Hah, didn't notice it was one expression!

thumphries:
Added a note, removed the `do`, and turned the dependencies into a recursive call to get rid of some redundant rebuilds

src/Mafia/Home.hs
```diff
@@ -51,30 +51,28 @@ installBinary package deps = do
...
+    createDirectoryIfMissing True prefix
+    renameFile (prefix <> ".tmp") (prefix </> (unPackageName . pkgName) package)
+  exists <- liftIO (doesDirectoryExist path)
+  unless exists $
+    EitherT . withTempDirectory (T.unpack tmp) (T.unpack $ nv <> ".") $ \sandboxDir -> runEitherT $ do
+      -- Build the binary and its dependencies in a new sandbox.
+      -- Sandbox necessary as binary's deps could clash with the local project
+      Pass <- callFrom id (T.pack sandboxDir) "cabal" ["sandbox", "init"]
+      -- Install any required executables first
+      for_ deps ensureExeOnPath
```
charleso: I guess in theory we could just make the consumers do this if it's not using the same sandbox (happy to do that separately though)

jacobstanley:
:8ball: looks great

thumphries:
Not sure why CI doesn't like this - looks like it can't find cabal or ghc or something

charleso:
`Downloading: Cabal-1.Test suite test-cli: FAIL` yeah something is very wrong.

charleso:
And the next build is fine

charleso:
@thumphries It might be that the test is shit and the `grep` I have for testing `hoogle` is picking up the install log output.

thumphries:
ok, ci is happy. I added a success marker file for each package, and a file lock around the install operation

@jystic @charleso could you take a glance at that second commit?

src/Mafia/Home.hs
```diff
@@ -51,30 +52,28 @@ installBinary package deps = do
...
-    renameFile path (path <> ".tmp")
-    createDirectoryIfMissing True path
-    renameFile (path <> ".tmp") (path </> (unPackageName . pkgName) package)
-  liftIO (doesDirectoryExist path) >>= \case
-    True ->
-      pure path
-    False -> do
+  whenM (liftIO (doesFileExist prefix)) (removeFile prefix)
+  bracketEitherT' (liftIO (lockFile lock Exclusive)) (liftIO . unlockFile) $ \_ ->
+    unlessM (liftIO (doesFileExist marker)) $
```
charleso, line 66: Should this go on the outside of the `bracket`?

charleso:
:+1: although just confirm with @jystic

src/Mafia/Home.hs
thumphries, line 66: I think it's in the right spot
- a rival build is running and fails, we wait for the lock, see no marker file, start a new build
- rival build succeeds, we get the lock, see marker file, no need to build

jacobstanley, line 66: I think you need one inside _and_ one outside

thumphries, line 66: to remove redundant locking? or is there a race i'm missing?

PR #47, opened 2016-01-08
Quick can now take multiple files as input

jacobstanley:
Usage:

```
mafia quick -i../disorder-core/src $(find src -name '*.hs')
```

^ I think with a bit of work we can make `mafia repl` do this automatically and at the same time prevent the source dependencies we know we will have loaded (in this case `disorder-core`) from being built.

/cc @markhibberd

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -67,7 +67,7 @@ data MafiaCommand
   | MafiaTestCI [Argument]
   | MafiaRepl   [Argument]
   | MafiaBench  [Argument]
-  | MafiaQuick  [GhciInclude] File
+  | MafiaQuick  [GhciInclude] [File]
   | MafiaWatch  [GhciInclude] File [Argument]
   | MafiaHoogle [Argument]
   deriving (Eq, Show)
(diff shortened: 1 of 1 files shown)


main/mafia.hs
```diff
@@ -234,37 +234,41 @@ bench args = do
...
-            <> [ path ]
+  return $ [ "-no-user-package-db" ]
+        <> (fmap ("-i" <>)           includes)
+        <> (fmap ("-package-db=" <>) databases)
+        <> paths
+
+checkEntryPoint :: File -> EitherT MafiaError IO ()
+checkEntryPoint file = do
+  exists <- doesFileExist file
+  case exists of
```
charleso: `unless` (or `unlessM` if you have it)? :) /nit

markhibberd:
For the non-bash non-savages:

```
mafia quick -i../disorder-core/src **/*.hs
```

main/mafia.hs
jacobstanley: oh yeah, `unlessM` is in P now I think

jacobstanley:
> For the non-bash non-savages:
> 
> ```
> mafia quick -i../disorder-core/src **/*.hs
> ```

wow, TIL

jacobstanley:
@charleso does this look ok now?

charleso:
:+1:

PR #48, opened 2016-01-11
Support different flavours of packages

jacobstanley:
This allows the profiling build or package documentation to be installed
when required, the package is "upgraded" in place in the global cache.

### DIFF
--- a/src/Mafia/Init.hs
+++ b/src/Mafia/Init.hs
@@ -78,7 +78,7 @@ initialize mprofiling = do
   firstEitherT InitGitError initSubmodules
 
   sandboxDir <- firstEitherT InitCabalError initSandbox
-  let statePath = sandboxDir </> "mafia/state.json"
+  let statePath = sandboxDir </> "mafia/state." <> T.pack (show mafiaStateVersion) <> ".json"
 
   clearAddSourceDependencies sandboxDir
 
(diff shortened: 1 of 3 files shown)


src/Mafia/Init.hs
```diff
@@ -150,18 +154,16 @@ data Profiling =
...
 
-data InstallState =
-  InstallState {
-      isSourceDependencies :: Set SourcePackage
-    , _isCabalFile         :: Hash
-    } deriving (Eq, Ord, Show)
-
 data MafiaState =
   MafiaState {
-      msInstallState :: InstallState
```
charleso, line 161: Any reason for this change apart from cleaning up? Looks like the json file has the same information?

src/Mafia/Install.hs
```diff
@@ -163,50 +181,96 @@ link w db env p@(Package (PackageRef pid _ _) _ _) = do
...
+
+flavourArgs :: Flavour -> [Argument]
+flavourArgs = \case
+  Vanilla ->
+    []
+  Profiling ->
+    [ "--reinstall"
+    , "--ghc-options=-fprof-auto-exported"
+    , "--enable-library-profiling" ]
+  Documentation ->
```
charleso, line 232: I guess at some point we'll want to call this again

charleso:
:+1:

src/Mafia/Init.hs
jacobstanley, line 161: yeah just cleaning up

jacobstanley, line 161: whether profiling is installed is morally part of the install state, which is just the mafia state now

bumping the version on the json state is basically free, recreating it only takes 5-10secs

src/Mafia/Install.hs
jacobstanley, line 232: yeah i think we'd call that when someone asks for hoogle or for haddock docs at some point

jacobstanley:
@charleso can you take a look at the new "Install hackage packages" commit

src/Mafia/Install.hs
```diff
@@ -163,50 +181,108 @@ link w db env p@(Package (PackageRef pid _ _) _ _) = do
...
+  ignoreIO (removeDirectoryRecursive sbdir)
+  createDirectoryIfMissing True sbdir
+
+  Hush <- sbcabal "sandbox" ["init", "--sandbox", sbdir]
+
+  case msrc of
+    -- hackage package
+    Nothing -> do
+      createDirectoryIfMissing True sbsrc
+      Hush <- sbcabal "unpack" ["--destdir=" <> sbsrc, renderPackageId pid]
```
charleso, line 265: A comment for why you do this might be good - it won't be obvious over time :)

charleso:
:+1:

PR #49, opened 2016-01-11
Additional options for low-level profiling

olorin:
/cc @jystic

### DIFF
--- a/src/Mafia/Init.hs
+++ b/src/Mafia/Init.hs
@@ -121,8 +121,9 @@ profilingArgs = \case
     ]
   EnableProfiling  -> [
       "--enable-profiling"
-    , "--ghc-options=-fprof-auto-top"
     , "--disable-executable-stripping"
+    , "--ghc-options=-fprof-auto-top"
+    , "--ghc-options=-eventlog"
     ]
 
 -- If a user or an older version of mafia has used 'cabal sandbox add-source'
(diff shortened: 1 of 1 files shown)


jacobstanley:
@olorin nice

Would there be a use for unstripped executables without profiling enabled? might be handy for profiling using linux-perf tools without the overhead, although i'm unsure

olorin:
@jystic possibly; it does seem like they could be used independently, though profiling with just `perf-events` is a lot harder than `perf-events` combined with GHC instrumentation. Do you think a separate flag would be a good idea?

jacobstanley:
@olorin yeah, I was thinking something like that, although, maybe we should just keep it simple and wait until we actually have a real need for it before we add another flag.

Let's merge this as is I think :cake:

jacobstanley:
Not sure if @markhibberd has any thoughts on stripping executables?

src/Mafia/Init.hs
```diff
@@ -115,8 +115,15 @@ profilingFlavour = \case
...
-  DisableProfiling -> ["--disable-profiling"]
-  EnableProfiling  -> ["--enable-profiling", "--ghc-options=-fprof-auto-top"]
+  DisableProfiling -> [
+      "--disable-profiling"
+    , "--enable-executable-stripping"
+    ]
+  EnableProfiling  -> [
+      "--enable-profiling"
+    , "--disable-executable-stripping"
+    , "--ghc-options=\"-fprof-auto-top -eventlog\""
```
jacobstanley: Looks like this needs to be:

``` haskell
, "--ghc-options=-fprof-auto-top"
, "--ghc-options=-eventlog"
```

olorin: Yup, fixed.

PR #50, opened 2016-01-12
Install binary broken due to missing directory

charleso:
/cc @thumphries @jystic 

The hoogle tests must pass because of the shared home directory :(

### DIFF
--- a/src/Mafia/Home.hs
+++ b/src/Mafia/Home.hs
@@ -62,6 +62,7 @@ installBinary package deps = do
   -- We can remove this at some point
   -- NOTE: This makes no attempt at being threadsafe
   whenM (liftIO (doesFileExist prefix)) (removeFile prefix)
+  createDirectoryIfMissing True prefix
   unlessM (liftIO (doesFileExist marker)) . withLock $
     unlessM (liftIO (doesFileExist marker)) $
       EitherT . withTempDirectory (T.unpack tmp) (T.unpack $ nv <> ".") $ \sandboxDir -> runEitherT $ do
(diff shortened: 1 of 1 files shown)


olorin:
:8ball:

thumphries:
weird. Is it the parent directory that's missing? I had no trouble deleting and reinstalling various things locally

olorin:
@thumphries looks to me like the current version will break on `withLock` when `prefix` doesn't exist.

charleso:
Yeah, the lock file failed for me: `.ambiata/mafia/bin/ghcid-0.5/lock: openFd: does not exist (No such file or directory)`

thumphries:
aaaack, added that lock too hastily. my bad. :lock:

src/Mafia/Home.hs
```diff
@@ -62,6 +62,7 @@ installBinary package deps = do
   -- We can remove this at some point
   -- NOTE: This makes no attempt at being threadsafe
   whenM (liftIO (doesFileExist prefix)) (removeFile prefix)
+  createDirectoryIfMissing True prefix
```
jacobstanley, line 65: Would this throw if there was a race and the directory already existed? I wouldn't normally worry about that for something like this, but we are going to the trouble of taking a lock.

jacobstanley, line 65: It would probably be safe if you handle/catch `IOException` and ignore it

thumphries, line 65: it doesn't throw if the dir already exists http://hackage.haskell.org/package/directory-1.2.5.0/docs/src/System.Directory.html#createDirectoryIfMissing

jacobstanley, line 65: Ignore me, @charleso pointed out this is safe against the race condition where two mafia's try to create the same directory
http://hackage.haskell.org/package/directory-1.2.5.0/docs/src/System.Directory.html#createDirectoryIfMissing

markhibberd, line 65: > Unfortunately there is a slight race condition here, but we _think_ it is benign.

:runs_away:

jacobstanley, line 65: > > Unfortunately there is a slight race condition here, but we _think_ it is benign.
> 
> :runs_away:

That race is technically possible in this case as well, because we used to create a file with the same name as the directory.

Very unlikely though :)

jacobstanley:
:+1:

PR #52, opened 2016-01-12
Add extra check for package flavour marker

jacobstanley:
Without this we end up doing an exponential number of checks, which can take a huge amount of additional time (> 30 seconds) for some of our more complicated projects.


charleso:
It was certainly very very slow, this looks like an improvement :+1:

PR #53, opened 2016-01-12
Only handle IOException

jacobstanley:
This was catching `ThreadKilled` and causing problems with the `async` library's cancellation mechanism.


charleso:
Surface :+1:

jacobstanley:
I think ideally we have something in P for catching all "safe" exceptions, but I'm happy with this for the moment as I think the `System.Process` stuff only throws `IOException`

PR #54, opened 2016-01-12
Added 'mafia clean'

jacobstanley:
This calls cabal clean and also deletes the sandbox

### DIFF
--- a/src/Mafia/Cabal/Sandbox.hs
+++ b/src/Mafia/Cabal/Sandbox.hs
@@ -4,6 +4,7 @@ module Mafia.Cabal.Sandbox
   ( sandbox
   , sandbox_
   , initSandbox
+  , removeSandbox
 
   , getSandboxDir
   , getPackageDB
@@ -43,11 +44,9 @@ sandbox_ cmd args = do
 
 -- Sandbox initialized if required, this should support sandboxes in parent
 -- directories.
-initSandbox :: EitherT CabalError IO Directory
+initSandbox :: EitherT CabalError IO SandboxDir
 initSandbox = do
-  ghcVer <- firstEitherT CabalGhcError getGhcVersion
-  let sandboxDir = ".cabal-sandbox" </> ghcVer
-
+  sandboxDir <- getInferredSandboxDir
   cfgSandboxDir <- liftIO (runEitherT getSandboxDir)
   let cfgOk = dropLeft cfgSandboxDir == Just sandboxDir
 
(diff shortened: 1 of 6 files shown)


tranma:
more like 'mafia nuke' 😄
:+1:

jacobstanley:
I'm reserving 'mafia nuke' for when you want to delete your global cache

tranma:
'mafia deathstar' for reinstalling ghc?

jacobstanley:
mafia defenestrate

test/Test/IO/Mafia/Chaos.hs
```diff
@@ -72,7 +74,8 @@ data Profiling =
     deriving (Eq, Ord, Show)
 
 data Action =
-    Build Profiling
+    Clean
```
charleso, line 77: Missing from `genAction`?

src/Mafia/Cabal/Sandbox.hs
```diff
@@ -62,17 +61,42 @@ dropLeft :: Either a b -> Maybe b
...
 
+removeSandbox :: EitherT CabalError IO ()
+removeSandbox = do
+  -- remove sandbox directory and sandbox config file
+  dir <- getInferredSandboxDir
+  ignoreIO $ removeDirectoryRecursive dir
+  ignoreIO $ removeFile defaultConfig
+
+  -- remove root sandbox directory if it is now empty
+  whenM (doesDirectoryExist sandboxRoot) . ignoreIO $ do
```
thumphries: why ignore? wouldn't we want to throw left / explode if this went wrong? likewise the two above

charleso:
:+1:

src/Mafia/Cabal/Sandbox.hs
jacobstanley: It might get removed by something else after the directory exists check

jacobstanley: Maybe I shouldn't bother protecting against that kind of thing here, just fail it happens

thumphries: should probably handle DNE explicitly, lots of weird IOExceptions out there that we should yell about

test/Test/IO/Mafia/Chaos.hs
jacobstanley, line 77: good catch, fixed

src/Mafia/Cabal/Sandbox.hs
jacobstanley: I might just make this whole function assume a single thread, the rest of cabal probably can't handle multiple builds at the same time anyway

thumphries: :+1:

PR #55, opened 2016-01-13
Use mktemp instead of /tmp/mafia

jacobstanley:
/cc @charleso @olorin 

The idea with this is we could invoke it at the start of a build so we have some idea which projects are up to date when looking at logs in jenkins.

I also noticed we're downloading to `/tmp/mafia` so I changed it to a `mktemp`, sorry that probably should be a separate PR.

### DIFF
--- a/script/mafia
+++ b/script/mafia
@@ -9,24 +9,40 @@ fetch_latest () {
   fi
 }
 
+latest_version () {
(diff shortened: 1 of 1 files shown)


olorin:
LGTM :+1:

jacobstanley:
As discussed IRL with @charleso this should probably live in anatomy instead, I'll fix it up so this just includes the /tmp fix

PR #56, opened 2016-01-13
Create MAFIA_BIN directory earlier in script

jacobstanley:
The script would fail if `~/.ambiata/mafia/bin` didn't exist yet as that is where it tries to create the temp file


nhibberd:
:zap:

thumphries:
:diamond_shape_with_a_dot_inside:

PR #57, opened 2016-01-14
Upgrade nitpicks

### LINKED ISSUE #22: upgrade does not respect symlinks

olorin:
- don't assume `mafia` is always `./mafia`
- bail out when trying to upgrade a symlink (fixes #22)
- don't print success message on upgrade until the new version is in place

/cc @jystic @nhibberd

### DIFF
--- a/script/mafia
+++ b/script/mafia
@@ -19,13 +19,19 @@ local_version () {
 
 run_upgrade () {
   MAFIA_TEMP=$(mktemp 2>/dev/null || mktemp -t 'upgrade_mafia')
-  MAFIA_CUR="$0"
 
   clean_up () {
     rm -f "$MAFIA_TEMP"
   }
 
   trap clean_up EXIT
+
+  MAFIA_CUR="$0"
+
+  if [ -L "$MAFIA_CUR" ]; then
+    echo 'Refusing to overwrite a symlink; run `upgrade` from the canonical path.' >&2
+    exit 1
+  fi
 
   echo "Checking for a new version of mafia ..."
   fetch_latest > $MAFIA_TEMP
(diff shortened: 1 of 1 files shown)


script/mafia
```diff
@@ -26,16 +26,23 @@ run_upgrade () {
 
   trap clean_up EXIT
 
+  MAFIA_CUR="$0"
+
+  if [ -L "$MAFIA_CUR" ]; then
+    echo 'Refusing to overwrite a symlink; run `upgrade` from the canonical path.' >&2
```
charleso, line 32: :100:

olorin, line 32: Tangential, anyone know a non-horrible portable way to dereference a symlink in shell? Best I can come up with is parsing `stat(1)` output.

jacobstanley:
:100: I was complaining about this behaviour this morning

script/mafia
jacobstanley, line 32: Is `readlink` portable?

olorin, line 32: The C function is, but I don't believe the command is part of POSIX. I've never used a UNIX that didn't have it though, so maybe it's good enough.

jacobstanley, line 32: It's available on OS/X so I'm happy with that.

Behaviour on OS/X is as follows (in case it differs from your setup):

```
file does not exist          = no output, exit code 1
file exists, but not symlink = no output, exit code 1
file exists, is symlink      = contents of symlink to stdout, exit code 0
```

markhibberd, line 32: portable readlink is hard because the default "not-a-link" behaviour is different in all the implementations (the is a link behaviour is consistent). I have a shell 1-liner that I have used in the past to do it that works across *BSD, linux, solaris, but in attempting to remember it I tried to log on to my server at home and it appears my internode connection is out :grapes:

markhibberd, line 32:
```
file exists, but not symlink = no output, exit code 1
```

That is the problematic case. On linoox it return 0 and prints canonical path if it is a directory.

PR #59, opened 2016-01-17
Allow mafia to be used with the development version (1.23.*) of cabal

jacobstanley:
cabal >= 1.23 is required for building with GHC 8.0 RC1, and thanks to recent commits, it now works properly with mafia


charleso:
We should probably add a regression test for supported version at some point. :+1:

jacobstanley:
Yeah that would be great

PR #62, opened 2016-01-18
Show a warning if submodules are not in sync with the git index

jacobstanley:
Not 100% sure about this one, putting up the PR because I'm looking for feedback.

Sometimes people encounter build errors because their submodules are not up to date and they need to `git submodule update`.

This PR shows a warning for each submodule which needs updating, so if p and x were pointing to a different commit than what the git index said they should, you would see this:

```
Warning: lib/p is out of sync with the git index
Warning: lib/x is out of sync with the git index
```

Some things to consider:
- Is this really mafia's responsibility?
- Is the message too verbose? (i.e. would it become annoying?)

### DIFF
--- a/src/Mafia/Git.hs
+++ b/src/Mafia/Git.hs
@@ -13,8 +13,11 @@ module Mafia.Git
     , getSubmodules
     ) where
 
+import           Control.Monad.IO.Class (liftIO)
+
 import           Data.Text (Text)
 import qualified Data.Text as T
+import qualified Data.Text.IO as T
 
 import           Mafia.Path
 import           Mafia.Process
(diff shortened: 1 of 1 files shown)


markhibberd:
Not sure, depends on what scenarios you are trying to catch - I assume it is just for interactive reminder if you have switched branches and not run update (as opposed to intentionally modifiying source)?

jacobstanley:
@markhibberd yeah that's the idea, I think it wouldn't come up if you've just modified the source, because the git hash of the submodule hasn't changed

if you commit the changes in the submodule then it would appear

markhibberd:
@jystic I think you should just merge and if it turns out to be noise, then we can revert. Don't think it is going to hurt anyway.

PR #65, opened 2016-01-28
firstEitherT -> firstT

jacobstanley:
Some formatting changes and switching to the new P which has `class BifunctorTrans`


tmcgilchrist:
looks good :+1:

PR #66, opened 2016-01-28
Add cabal configure flags support

jacobstanley:
Usage:

``` shell
$ mafia build -fmyflag   # myflag on
$ mafia build -f+myflag  # myflag on
$ mafia build -f-myflag  # myflag off
```

This is to fix https://github.com/ambiata/ambiata-cli/pull/39


markhibberd:
:+1:

PR #67, opened 2016-01-29
Use temporary file to store sdist source list

jacobstanley:
This also puts `withSystemTempDirectory` in the `Mafia.IO` module, replacing all the usages of the `[Char]` variant.

/cc @markhibberd


markhibberd:
:+1:

PR #70, opened 2016-02-09
Accidentally O(scary)

jacobstanley:
Sorry everyone, I'm an idiot :(

The hashes will change in this version because I'm now storing package dependencies in a `Set Package` like I should have been doing all along. I think I didn't do this originally because I was just using the default derived instance of `Ord` for `Package` which needed to compare all of a packages dependencies as part of the ordering. Now I have a custom instance which only compares the `PackageRef` and the `Hash`.

This improves the time to calculate transitive dependencies on `hydra-daemon` from ~20-30 mins -> 16 seconds on my machine.

The performance improvements include:
- Not constructing a `Data.CaseInsenitive.CI` for every `Ord` comparison (biggest win)
- Keep packages stored in a `Set` so we don't need to keep reconstructing it to do unions
- Parallel map
- Strictify/unpack fields

/cc @markhibberd

### DIFF
--- a/ambiata-mafia.cabal
+++ b/ambiata-mafia.cabal
@@ -29,6 +29,7 @@ library
                     , exceptions                      == 0.8.*
                     , filelock                        == 0.1.*
                     , filepath                        == 1.4.*
+                    , parallel                        == 3.2.*
                     , process                         == 1.2.*
                     , tar                             == 0.4.*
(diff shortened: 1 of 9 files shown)


markhibberd:
:+1: Thanks man.

tranma:
good to know :+1:

PR #73, opened 2016-02-13
Use the newest available installed mafia to bootstrap

jacobstanley:
This changes the bootstrap script used when building a version of mafia you don't have yet.

The bootstrap script will now find the latest available version of mafia, and use that to build instead of cabal, so that we can benefit from the global cache.

If for some reason this fails, it will do a `git clean -xdf` and try the old cabal method of building.

The contract we are assuming here, is that for all mafia versions, `mafia upgrade && mafia build` will cause an executable to be produced at `dist/build/mafia/mafia`.

/cc @markhibberd @charleso

### DIFF
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -1,13 +1,53 @@
 #!/bin/sh -eu
 
-git submodule update --init
+#
+# List the version and path to all the installed versions of mafia.
+#
+installed_mafia_versions () {
+  for MAFIA in $(find $HOME/.ambiata/mafia/bin -maxdepth 1 -type f -name 'mafia-*' -exec test -x {} \; -print); do
+    MAFIA_VERSION=$($MAFIA --version)
+    echo "$MAFIA_VERSION $MAFIA"
+  done
+}
 
-cabal sandbox init
+#
+# Find the latest version of mafia which is currently installed.
+#
+latest_mafia () {
+  LATEST=$(installed_mafia_versions | sort --reverse | head -1 | cut -d ' ' -f 2)
+  if [ -z "$LATEST" ]; then
+    return 1
+  else
+    echo "$LATEST"
+  fi
+}
 
-CABAL_SOURCES=$(find lib -maxdepth 4 ! -path lib/\*/bin/\* ! -path lib/\*/lib/\* -name \*.cabal | xargs -L 1 dirname)
-for CABAL_SOURCE in $CABAL_SOURCES; do
-  cabal sandbox add-source -- $CABAL_SOURCE
-done
+#
+# Build using the latest installed version of mafia.
+#
+mafia_build () {
+  MAFIA=$(latest_mafia) || return 1
+  $MAFIA update
+  $MAFIA build
+}
 
-cabal update
-cabal install
+#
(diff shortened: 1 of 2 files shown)


tmcgilchrist:
@jystic curious, what situation are you left in if `mafia upgrade && mafia build` fails in the middle? Assume you can just run it again. Can it get properly broken at that point?

jacobstanley:
@tmcgilchrist the way the script works, it would do a `git clean -xdf` and then fall back to the old method of using `cabal` to install it.

script/mafia
```diff
@@ -87,7 +87,7 @@ exec_mafia () {
 
         bin/bootstrap ) || exit $?
 
-      cp "$MAFIA_TEMP/mafia/.cabal-sandbox/bin/mafia" "$MAFIA_PATH_TEMP"
+      cp "$MAFIA_TEMP/mafia/dist/build/mafia/mafia" "$MAFIA_PATH_TEMP"
```
thumphries, line 90: Could this be `dist/build/Mafia` on some machines, like case-sensitive HFS+? /notsure

tmcgilchrist, line 90: who'd be crazy enough to do that?

also linux might see issues with that path

jacobstanley, line 90: there should be two directories "dist/build/mafia" and "dist/build/Mafia", not sure how case sensitive HFS+ works with that? with case insensitive it thinks they're the same and dumps all the files in there like a savage

thumphries, line 90: hahaha yeah that's why I was confused - didn't know cabal was racing itself on case-insensitive filesystems

bin/bootstrap
```diff
@@ -1,13 +1,53 @@
...
+#
+# Build using cabal.
+#
+cabal_build () {
+  git clean -xdf
+  git submodule update --init
+
+  cabal sandbox init
+
+  CABAL_SOURCES=$(find lib -maxdepth 4 ! -path lib/\*/bin/\* ! -path lib/\*/lib/\* -name \*.cabal | xargs -L 1 dirname)
```
thumphries, line 43: brvtal

thumphries:
:shaved_ice:

tmcgilchrist:
:rocket:

jacobstanley:
I'm not 100% sure about merging this one, haven't convinced @charleso and @markhibberd it's a good idea yet :)

The only reasonable alternative is OS/X build bots making binaries. @markhibberd any idea what it would take to hook jenkins up to that?

markhibberd:
> I'm not 100% sure about merging this one, haven't convinced @charleso and @markhibberd it's a good idea yet :)

My main hesitation is as mafia stabilises is this worth it? This currently costs maybe half an hour a week, but I feel like that will drop away pretty dramatically over time. Not sure.

> The only reasonable alternative is OS/X build bots making binaries. @markhibberd any idea what it would take to hook jenkins up to that?

Probably not much, maybe a half day of stuffing around with infrastructure.

markhibberd:
Decision? @jystic

jacobstanley:
@markhibberd I think this is a no for the moment - the right way is OS/X build bots.

PR #75, opened 2016-02-16
mafia install <package>

jacobstanley:
This allows the use of mafia's global cache for installing executables from Hackage.

Usage:

```
$ mafia install pretty-show
stderr> Building...
stderr> Building...
stdout> /Users/jdoe/.ambiata/mafia/bin/pretty-show/bin
```

Or even better:

```
$ echo "Foo(Bar     (Baz 1 2    3)   )" | $(mafia install pretty-show)/ppsh
Foo (Bar (Baz 1 2 3))
```

It also works with a version number suffix:

```
$ echo "[Max\n1,ML\n2]" | $(mafia install pretty-show-1.6.9)/ppsh
[ Max 1, ML 2 ]
```

The existing `MAFIA_HAPPY`, `MAFIA_ALEX` things now use the global cache as a result of this change.

### DIFF
--- a/src/Mafia/Bin.hs
+++ b/src/Mafia/Bin.hs
@@ -1,0 +1,52 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
+module Mafia.Bin
+  ( installBinary
+  , ensureExeOnPath
+  ) where
+
+import           Control.Monad.IO.Class (MonadIO(..))
+
+import qualified Data.Text as T
+
+import           Mafia.Home
+import           Mafia.IO
+import           Mafia.Install
+import           Mafia.Package
+import           Mafia.Path
+import           P
+
+import           System.IO (IO)
+import           System.Posix.Files (createSymbolicLink)
+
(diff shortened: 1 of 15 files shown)


tmcgilchrist:
:+1: looks really useful.

src/Mafia/Bin.hs
```diff
@@ -0,0 +1,89 @@
...
+    env <- getPackageEnv
+    let gdir = packageSandboxDir env pkg
+    liftIO $ createSymbolicLink (T.unpack gdir) (T.unpack plink)
+
+  return pbin
+
+ensureExeOnPath :: InstallPackage -> EitherT InstallError IO ()
+ensureExeOnPath pkg = do
+  dir <- installBinary pkg
+  setEnv "PATH" . maybe dir (\path -> dir <> ":" <> path) =<< lookupEnv "PATH"
```
charleso, line 109: I know I added this, but I wish I'd done something a little safer - like creating a `Exe` data type which you pass around.

charleso:
:+1:

src/Mafia/Bin.hs
jacobstanley, line 109: so you just put it on the path temporarily?

jacobstanley:
I noticed you'd have a strange situation if you install `async` for example. I added a small patch to show an error that the bin/ directory cannot be linked as there are no executables. The package will still be installed however.

PR #81, opened 2016-02-17
Remove the need for a .submodules file in all sensible cases

jacobstanley:
This patch removes the need for a `.submodules` file completely. Any `.cabal` projects in the current repository are now made available as source dependencies.

I have left in support for the `.submodules` file as it can still be useful if you want to temporarily reference something from outside your git repo, but I don't think it needs to be documented in the high level readme.

/cc @tmcgilchrist

### DIFF
--- a/src/Mafia/Submodule.hs
+++ b/src/Mafia/Submodule.hs
@@ -103,9 +103,16 @@ getConfiguredSources = do
 
 getConventionSources :: EitherT GitError IO (Set Directory)
 getConventionSources = do
+  -- make sure we never include the current directory (i.e. the project we're
+  -- trying to build) as a source dependency
+  dir <- getCurrentDirectory
   root <- getProjectRoot
+  repoPaths <- Set.filter (/= dir) <$> getSourcesFrom root
+
   subs <- fmap ((root </>) . subName) <$> getSubmodules
-  Set.unions <$> traverse getSourcesFrom subs
+  subPaths <- Set.unions <$> traverse getSourcesFrom subs
+
+  return $ Set.union repoPaths subPaths
 
 getSourcesFrom :: MonadIO m => Directory -> m (Set Path)
 getSourcesFrom dir = do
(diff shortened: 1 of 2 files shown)


tranma:
if it's still possible to use a .submodules files for savage reasons, maybe mention in the docs?

jacobstanley:
> if it's still possible to use a .submodules files for savage reasons, maybe mention in the docs?

Maybe i should rename it to `.savmodules`

src/Mafia/Submodule.hs
```diff
@@ -103,9 +103,16 @@ getConfiguredSources = do
...
+  -- trying to build) as a source dependency
+  dir <- getCurrentDirectory
   root <- getProjectRoot
+  repoPaths <- Set.filter (/= dir) <$> getSourcesFrom root
+
   subs <- fmap ((root </>) . subName) <$> getSubmodules
-  Set.unions <$> traverse getSourcesFrom subs
+  subPaths <- Set.unions <$> traverse getSourcesFrom subs
+
+  return $ Set.union repoPaths subPaths
```
markhibberd, line 115: Untested, but what happens when there are ordering constraints, is this filtered by actual deps at some point?

i.e. 

```
  project-core
  project-daemon
  project-http
  project-cli
```

When project-core add-source on project-cli that has a dependency on project-core. Concern or not?

jacobstanley, line 115: Cabal figures it out, this just makes them available as packages, cabal only produces a build plan containing packages that it needs.

jacobstanley, line 115: If you try to make a circular dependency you get a relevant error message from cabal:

```
Checking for changes to dependencies...
Installing dependencies...
Resolving dependencies...
cabal: internal error: could not construct a valid install plan.
The proposed (invalid) plan contained the following problems:
The following packages are involved in a dependency cycle foo-0.0.1, bar-0.0.1, bar-test-0.0.1
-- snip --
```

jacobstanley:
> if it's still possible to use a .submodules files for savage reasons, maybe mention in the docs?

I agree it should probably be documented somewhere, but I'm not really sure where it fits in the readme. I was trying to make describing the getting started process a bit simpler by removing the need for this.

charleso:
:+1:

tmcgilchrist:
Great :thumbsup:

PR #82, opened 2016-02-18
Remove godfather image

jacobstanley:
This godfather image was cluttering the readme a bit, so I think maybe we should just keep the lego one for now.

Would bring it back if there was an good place to put it.

/cc @damncabbage


thumphries:
:chart_with_downwards_trend: rip

PR #84, opened 2016-02-21
Add 'mafia lock' for locking down package versions

jacobstanley:
This adds two new commands, `mafia lock` and `mafia unlock`.

`mafia lock` creates a `<project>.lock` file alongside the `<project>.cabal` file which contains the version numbers and flags of all the Hackage packages which were used during the last build. Subsequent builds add all of these constraints to any solving they do. This means that if someone releases a newer version of a locked down package on Hackage, then the newer version will be ignored.

Running `mafia lock` when you already have a `.lock` file can be used to update the `.lock` with any new packages you've added to your `.cabal` file without losing a lock on the existing ones.

`mafia unlock` simply deletes the `.lock` file, allowing Cabal's solver free reign over package choice.

I'm not completely sure whether lock files will be a net positive or negative, would be great to hear peoples thoughts? It certainly could be useful to avoid rebuilds at the present time, given that people are madly upgrading all their packages to support GHC 8.0.

Happy for people to bikeshed the file names and the way it it is invoked. Some package manager / build systems (such as Rust's Cargo) create a lock file automatically when you build. 

For the record, I tried to get Cabal's freeze/unfreeze stuff to work, but there's something about the way we do dependency solving in Mafia that means Cabal just ignores the freeze file which it creates, _shrugs_.

/cc @markhibberd @charleso @thumphries @damncabbage

### DIFF
--- a/src/Mafia/Init.hs
+++ b/src/Mafia/Init.hs
@@ -91,8 +91,11 @@ initialize mprofiling mflags = do
 
   clearAddSourceDependencies sandboxDir
 
+  defaultLockFile <- firstT InitLockError $ getLockFile =<< getCurrentDirectory
+  lockFile <- fromMaybe defaultLockFile <$> lookupEnv "MAFIA_LOCK"
+
   previous <- readMafiaState statePath
-  current <- getMafiaState (mprofiling <|> fmap msProfiling previous) (mflags <|> fmap msFlags previous)
+  current <- getMafiaState (mprofiling <|> fmap msProfiling previous) (mflags <|> fmap msFlags previous) lockFile
   hasDist <- liftIO $ doesDirectoryExist "dist"
 
   packages <- checkPackages
@@ -107,8 +110,11 @@ initialize mprofiling mflags = do
       flavours = profilingFlavour (msProfiling current)
       flags = msFlags current
 
-    lockFile <- firstT InitLockError $ getLockFile =<< getCurrentDirectory
-    constraints <- fmap (fromMaybe []) . firstT InitLockError $ readLockFile lockFile
+    constraints <-
+      if T.null lockFile then
+        pure []
+      else
+        fmap (fromMaybe []) . firstT InitLockError $ readLockFile lockFile
 
     installed <- firstT InitInstallError $ installDependencies flavours flags sdeps constraints
 
@@ -264,12 +270,12 @@ parseFlags :: Alternative f => [Text] -> f [Flag]
 parseFlags =
   traverse (maybe empty pure . parseFlag)
 
-getMafiaState :: Maybe Profiling -> Maybe [Flag] -> EitherT InitError IO MafiaState
-getMafiaState mprofiling mflags = do
+getMafiaState :: Maybe Profiling -> Maybe [Flag] -> File -> EitherT InitError IO MafiaState
+getMafiaState mprofiling mflags lockFile = do
   sdeps <- getSourceDependencies
   dir <- getCurrentDirectory
   chash <- getCabalFileHash dir
-  lhash <- getLockFileHash dir
+  lhash <- firstT InitHashError $ tryHashFile lockFile
 
   let
     profiling = fromMaybe DisableProfiling mprofiling
@@ -285,12 +291,6 @@ getCabalFileHash dir = do
       left (InitCabalFileNotFound dir)
     Just file ->
       firstT InitHashError (hashFile file)
-
-getLockFileHash :: Directory -> EitherT InitError IO (Maybe Hash)
-getLockFileHash dir = do
-  file <- firstT InitLockError $ getLockFile dir
-  mbytes <- readBytes file
-  pure $ fmap hashBytes mbytes
 
 getSourceDependencies :: EitherT InitError IO (Set SourcePackage)
 getSourceDependencies = do
--- a/src/Mafia/Init.hs
+++ b/src/Mafia/Init.hs
@@ -8,6 +8,8 @@ module Mafia.Init
   , initialize
   , getSourceDependencies
 
+  , readInstallConstraints
+
   , InitError(..)
   , renderInitError
   ) where
@@ -28,6 +30,7 @@ import           Mafia.Git
 import           Mafia.Hash
 import           Mafia.IO
 import           Mafia.Install
+import           Mafia.Lock
 import           Mafia.Path
 import           Mafia.Process
 import           Mafia.Submodule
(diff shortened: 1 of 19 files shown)


jacobstanley:
For an idea of what the `.lock` file looks like:

```
~/src/mismi/mismi-s3 $ mafia lock
~/src/mismi/mismi-s3 $ cat mismi-s3.lock
abstract-deque == 0.3
abstract-par == 0.3.3
adjunctions == 4.2.2
aeson == 0.8.0.2
aeson +old-locale
amazonka == 1.3.6
amazonka-core == 1.3.6
amazonka-core +old-locale
amazonka-s3 == 1.3.6
ansi-terminal == 0.6.2.3
...
```

markhibberd:
What does it do with submodule dependencies?

jacobstanley:
I figured they are already locked down by git, so I omitted them from the lock file.

markhibberd:
@jystic That makes it more useful then freeze. I am a bit out of date on current state of freeze, but the other problems that freeze had last time I looked (and I would be interested in how this stacks up):
- freeze across multiple ghc versions can cause problems
- freeze in libraries is painful because it applied recursively which meant resolution on large dependency trees extremely problematic

jacobstanley:
> - freeze across multiple ghc versions can cause problems

That's a point I hadn't considered, we should probably create `<project>.lock-<ghcver>` files instead.

> - freeze in libraries is painful because it applied recursively which meant resolution on large dependency trees extremely problematic

Are you saying that with submodules it would read the freeze file (e.g. `cabal.config`) from the submodule as well? So that if we had a freeze file for P, it would affect dependency resolution for everything?

Just to be clear about what I said earlier:
- the Mafia lock file includes all transitive Hackage dependencies
- lock files in source dependencies (such as submodules) have no effect on the current project
- source dependency version numbers are not included in the lock file as the version is implied

It's possibly worth including the flags chosen for source dependencies in the lock file.

As a side note, Mafia was already adding additional constraints during solving which force Cabal to pick the source package, even if it is available on Hackage as well.

jacobstanley:
I committed a lock file to mismi for giggles, then did a `cabal update; mafia unlock; mafia lock`, was interesting to see the change:

``` diff
diff --git a/mismi-s3/mismi-s3.lock-7.8.4 b/mismi-s3/mismi-s3.lock-7.8.4
index 28a4c49..32eb589 100644
--- a/mismi-s3/mismi-s3.lock-7.8.4
+++ b/mismi-s3/mismi-s3.lock-7.8.4
@@ -31,7 +31,7 @@ contravariant == 1.4
 cookie == 0.4.1.6
 criterion == 1.1.0.0
 cryptohash == 0.11.6
-cryptonite == 0.11
+cryptonite == 0.12
 data-default == 0.5.3
 data-default-class == 0.0.1
 data-default-instances-base == 0.0.1
@@ -65,7 +65,7 @@ monad-loops == 0.4.3
 monad-par == 0.3.4.7
 monad-par-extras == 0.3.3
 mtl == 2.2.1
-mwc-random == 0.13.3.2
+mwc-random == 0.13.4.0
 nats == 1.1
 network == 2.6.2.1
 network-info == 0.2.0.8
```

markhibberd:
All sounds pretty good to me.

damncabbage:
Something for later: it'd be nice to have an `upgrade <package>` where, to upgrade a package to a newer version, it filters out that package from the lockfile when handing the constraints over to Cabal, and then writes down whatever valid version comes back from the install. (I _think_ this is possible, unless I've misunderstood how this works. Definitely a separate chunk of work, to be clear.)

(I'd love to thumbs-up, but I don't have a good enough handle on how this all works. Excited all the same. :sweat_smile:)

jacobstanley:
> it'd be nice to have an `upgrade <package>`

Yeah it's a shame `mafia upgrade` is already in use as a command for bootstrapping script :/ But yes, something like that would be pretty awesome!

jacobstanley:
> it filters out that package from the lockfile when handing the constraints over to Cabal, and then writes down whatever valid version comes back from the install.

Until something like that is implemented, you can easily do it by hand. If you remove a package from the lock file, then run `mafia lock` again, it will re-solve with the modified lock file and produce a new lock file which has the (possibly) updated package version.

damncabbage:
Nice! Good to know.

Nowish would possibly be a good time to rename that to `self-upgrade` / `selfupgrade`, before this tool gets too much more public.

thumphries:
The feature and code looks good to me. Not sure where we'd use it in our infrastructure, but I'd probably use it locally.

Automatically generating the lock file on a clean build is quite interesting. It seems benign, but could introduce a gap between branches and master builds?

I can also imagine a world of hurt if we started committing lock files to repos (a build could be green on CI thanks to the lock, and fail hard when sourced as a submodule)

src/Mafia/Init.hs
```diff
@@ -205,6 +246,7 @@ instance FromJSON MafiaState where
       MafiaState <$>
         o .: "source-dependencies" <*>
         o .: "cabal-file" <*>
+        o .: "lock-file" <*>
```
charleso, line 263: Does this change the versioning in some way? It's mandatory by the looks.

charleso:
I think having an environment `MAFIA_LOCK_IGNORE` flag (or thereabouts) might help if we actually wanted to use this, and we could have an extra `canary` build with the locks ignored.

charleso:
Seems legit. Wouldn't mind discussing whether it's something we actually want to use in anger, or whether we prefer the current situation, and maybe discuss the pros/cons of each approach.

src/Mafia/Init.hs
charleso, line 263: > Does this change the versioning in some way? It's mandatory by the looks.

Whoops. Missed the version bump above. Do we support old versions as well?

jacobstanley, line 263: > Whoops. Missed the version bump above. Do we support old versions as well?

Nope, we don't need to, it's not compulsory information, just saves 5-10 seconds on a `mafia build`

jacobstanley:
> I can also imagine a world of hurt if we started committing lock files to repos (a build could be green on CI thanks to the lock, and fail hard when sourced as a submodule)

Yeah I think that's a problem in general that we'll have to experiment with, perhaps the canary builds would solve it.

jacobstanley:
> I think having an environment `MAFIA_LOCK_IGNORE` flag (or thereabouts) might help if we actually wanted to use this, and we could have an extra `canary` build with the locks ignored.

I added a `MAFIA_LOCK` environment variable that lets you specify which lock file to use. You can use `MAFIA_LOCK="" mafia build` to build without using a lock file.

More interestingly you could say `MAFIA_LOCK=lib/mismi/mismi-s3/mismi-s3.lock-7.8.4 mafia build` to use the lock file from another project. Running `mafia lock` always uses the constraints from the last build if it can, so you'd be able to save the versions which you used when building with the other project's lock file.

markhibberd:
:+1:

PR #85, opened 2016-02-22
Fix issue with installing happy/alex

jacobstanley:


src/Mafia/Install.hs
```diff
@@ -362,6 +388,34 @@ createPackageSandbox penv p@(Package (PackageRef pid _ msrc) deps _) = do
 
   return ty
 
+jenkins :: Directory -> Text
```
nhibberd: :sos:

src/Mafia/Install.hs
```diff
@@ -362,6 +388,34 @@ createPackageSandbox penv p@(Package (PackageRef pid _ msrc) deps _) = do
...
+        hash2 = hash1 `xor` (shiftR hash1 11)
+        hash3 = hash2 + (shiftL hash2 15)
+      in
+        hash3
+  in
+    T.pack .
+    flip showHex "" .
+    loop_finish .
+    foldl' loop 0 .
+    T.unpack
```
jacobstanley: This function is straight from Cabal, cleaned up to match our style a bit more

markhibberd: Maybe link to the original for prosperity.

markhibberd:
:wind_chime:

nhibberd:
:100:

PR #89, opened 2016-03-15
Add -rtsopts to profiling builds

olorin:
The one I want is `-M` (for debugging heap alloc issues), not aware of any way to enable it on its own - building rtsopts into a profiling build seems pretty harmless.

/cc @jystic

### DIFF
--- a/src/Mafia/Init.hs
+++ b/src/Mafia/Init.hs
@@ -154,6 +154,7 @@ profilingArgs = \case
       "--enable-profiling"
     , "--disable-executable-stripping"
     , "--ghc-options=-fprof-auto-top"
+    , "--ghc-options=-rtsopts"
     ]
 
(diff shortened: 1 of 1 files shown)


olorin:
Forgot @jystic was away, anyone else want to review? /cc @ambiata/engineers

amosr:
looks good to me. would there be any disadvantage to enabling it for all builds?
:rotating_light:

olorin:
None that I know of, just figured there was probably a reason it wasn't there - @nhibberd do you know?

thumphries:
Does -rtsopts only affect executables, or does it impose a change on libraries also?

olorin:
Only executables AFAIK.

thumphries:
cool, was worried it'd invalidate the mafia cache

agree its harmless and useful, I might be missing something though

nhibberd:
Given that profiling is now an option, cant see any reason not to

PR #91, opened 2016-03-18
Include -rtsopts in all builds

olorin:
I'm not aware of any downside to doing this, and it'd be helpful for testing for me (I'd like to add some tests which verify a thing runs in constant memory by running it with a tiny heap).

/cc @ambiata/engineers

### DIFF
--- a/src/Mafia/Init.hs
+++ b/src/Mafia/Init.hs
@@ -134,7 +134,9 @@ initialize mprofiling mflags = do
       profilingArgs (msProfiling current) <>
       fmap flagArg (msFlags current) <>
       [ "--enable-tests"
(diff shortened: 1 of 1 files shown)


markhibberd:
@olorin Why not in the cabal file? feels a bit funny having the default build options tied to mafia rather than the cabal file. As an alternative is it enough to say the default in our cabal files is to have this set. Or am I understating the importance of this, as in is it never valid to not have it set and is valid for mafia to be more opinionated than ghc in this case?

olorin:
@markhibberd you're right, there's no reason to put this in mafia rather than in cabal files, not sure what I was thinking.

PR #97, opened 2016-04-04
Resolve symbolic links when looking up MAFIA_HOME

jacobstanley:
Cabal always uses absolute paths with symlinks resolved when referring to files, so we need to do the same.

This should hopefully fix the bug where we calculate the wrong jenkins hash for the sandbox directory when trying to build happy in a mafia cache that is symlinked to another path.

/cc @nhibberd @markhibberd


olorin:
Seems legit :+1:

PR #100, opened 2016-04-11
Switch away from deprecated cryptohash package

jacobstanley:
All the crypto stuff is now in `cryptonite`, depending on `cryptohash` now gives errors like this:

```
src/Mafia/Hoogle.hs:13:18:
    Ambiguous module name ‘Crypto.Hash’:
      it was found in multiple packages:
      cryptonite-0.13 cryptohash-0.11.9
```

/cc @charleso

### DIFF
--- a/src/Mafia/Hash.hs
+++ b/src/Mafia/Hash.hs
@@ -18,11 +18,13 @@ module Mafia.Hash
   , tryHashFile
   ) where
 
-import qualified Crypto.Hash.SHA1 as SHA1
+import           Crypto.Hash (Digest, SHA1)
+import qualified Crypto.Hash as Hash
 
 import           Data.Aeson (Value(..), ToJSON(..), FromJSON(..))
+import           Data.ByteArray (ByteArrayAccess, convert)
+import           Data.ByteArray.Encoding (Base(..), convertToBase)
 import           Data.ByteString (ByteString)
-import qualified Data.ByteString as B
 import qualified Data.ByteString.Base16 as Base16
 import qualified Data.ByteString.Lazy as L
 import qualified Data.Text as T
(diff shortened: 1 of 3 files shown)


olorin:
:8ball:

thumphries:
⛳

jacobstanley:
Turns out this is a little more complicated than I thought, but fixes for the build coming soon

jacobstanley:
I think this should be right now /cc @olorin @thumphries

nhibberd:
Have just used package imports in the past for those versions

jacobstanley:
@nhibberd the problem is that both packages are from the same author and `cryptohash` is now official [deprecated](http://hackage.haskell.org/package/cryptohash), so we might as well switch to the legit version.

From the above link:

> DEPRECATED: this library is still fully functional, but please use cryptonite for new projects and convert old one to use cryptonite. This is where things are at nowadays.

charleso:
:+1:

src/Mafia/Hash.hs
```diff
@@ -52,25 +54,25 @@ renderHashError = \case
...
 parseHash :: Text -> Maybe Hash
 parseHash hex =
-  let (bs, _) = Base16.decode (T.encodeUtf8 hex)
-  in if B.length bs == 20
-     then Just (Hash bs)
-     else Nothing
+  let
+    (bs, _) = Base16.decode (T.encodeUtf8 hex)
+  in
+    Hash <$> Hash.digestFromByteString bs
```
thumphries: length check no longer necessary?

jacobstanley: It is now done by [digestFromByteString](http://hackage.haskell.org/package/cryptonite-0.15/docs/Crypto-Hash.html#v:digestFromByteString)

> Try to transform a bytearray into a Digest of specific algorithm.
> If the digest is not the right size for the algorithm specified, then Nothing is returned.

thumphries: sweet 📧

olorin:
:+1: assuming length check is no longer necessary

src/Mafia/Hash.hs
jacobstanley: Gah, just tried it out, the docs are a lie :/ If the length is 21 then it still succeeds, :savage:

jacobstanley: Ah, actually, it's the `Base16.decode` causing the problem

PR #101, opened 2016-04-11
Add -w flag to disable warnings during a build

jacobstanley:
/cc @tranma @thumphries 

Usage

```
$ mafia build -w
```

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -68,7 +68,7 @@ data MafiaCommand =
   | MafiaHash
   | MafiaDepends DependsUI (Maybe PackageName) [Flag]
   | MafiaClean
-  | MafiaBuild Profiling [Flag] [Argument]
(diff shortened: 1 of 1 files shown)


tranma:
:+1:

thumphries:
should this silence them entirely or just turn off `-Werror`?

tranma:
I personally don't mind whether it's ghc -w or -Wwarn (since I'll do a normal build later)

thumphries:
-Wwarn / -Wall is my preference. either way, I'm glad to replace my `sorry` alias 👍

jacobstanley:
I matched it to the ghc `-w` argument i suppose, which silences them entirely

I'd be happy enough with either option, but I thought it might be weird to have a different meaning for `-w` between ghc and mafia.

thumphries:
oh true, consistent with ghc is good 🎴

PR #104, opened 2016-05-11
Has `mafia install` respect the MAFIA_(ALEX|HAPPY|CPPHS)=true flags.

damncabbage:
eg. `env MAFIA_ALEX=true MAFIA_HAPPY=true mafia install purescript`

(Paging @jystic.)

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -394,7 +394,8 @@ mafiaHoogle args = do
   hoogle hkg args
 
 mafiaInstall :: InstallPackage -> EitherT MafiaError IO ()
-mafiaInstall ipkg =
+mafiaInstall ipkg = do
+  ensureAuxillaryExes
   liftIO . T.putStrLn =<< firstT MafiaBinError (installBinary ipkg)
 
 ghciArgs :: [GhciInclude] -> [File] -> EitherT MafiaError IO [Argument]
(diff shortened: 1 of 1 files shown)


damncabbage:
PS:

sorry

thumphries:
Looks good 👲 

Wish we didn't need these env variables

damncabbage:
Thanks. Build's still not going, despite a manual kickoff. :\

thumphries:
ask boris?

damncabbage:
why always boris

(Thanks; passing build, and works locally. I really don't how to write a test for this one, sorry.)

main/mafia.hs
```diff
@@ -395,7 +395,9 @@ mafiaHoogle args = do
 
 mafiaInstall :: InstallPackage -> EitherT MafiaError IO ()
 mafiaInstall ipkg =
-  liftIO . T.putStrLn =<< firstT MafiaBinError (installBinary ipkg)
+  liftIO . T.putStrLn =<< do
+    ensureAuxillaryExes
+    firstT MafiaBinError (installBinary ipkg)
```
jacobstanley: This is fine, but it reads a little weird, why not?

``` hs
mafiaInstall ipkg = do
  ensureAuxillaryExes
  liftIO . T.putStrLn =<< firstT MafiaBinError (installBinary ipkg)
```

jacobstanley:
Look good to me 👍 

> Wish we didn't need these env variables

It should be possible for us to do this automatically as required, we just need to be able to read the `build-tools` section of the `.cabal` file of the package we're trying to install - maybe it's something for next yak day.

thumphries:
> maybe it's something for next yak day

made an issue #105 , one day

jacobstanley:
It would be great if we could do that, then we could use it for internal tools like the `sardine-compiler`, as it stands I just commit the generated file to make life easier.

main/mafia.hs
damncabbage: Rearranged; thanks.

PR #110, opened 2016-06-01
Build mafia using GHC 8.0

jacobstanley:
The tests for this depend on x being upgraded to ghc 8.0

/cc @tmcgilchrist @thumphries @erikd-ambiata

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 
 import           BuildInfo_ambiata_mafia
+import           DependencyInfo_ambiata_mafia
 
 import           Control.Concurrent (setNumCapabilities)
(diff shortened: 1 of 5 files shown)


main/mafia.hs
```diff
@@ -56,6 +57,8 @@ main = do
   dispatch parser >>= \case
     VersionCommand ->
       putStrLn buildInfoVersion >> exitSuccess
+    DependencyCommand ->
+      traverse putStrLn dependencyInfo >> exitSuccess
```
tmcgilchrist, line 61: Do you need to add `gen` to the `hs-source-dirs` for this executable?

tmcgilchrist, line 61: Looks like it's already there, never mind.

tmcgilchrist:
Once 8.0.1 build is good 🌽

thumphries:
lgtm 🍧

PR #116, opened 2016-06-27
boris-git.toml

jacobstanley:
/cc @nhibberd @thumphries

### DIFF
--- a/boris-git.toml
+++ b/boris-git.toml
@@ -1,8 +1,26 @@
 [boris]
   version = 1
 
-[build.dist]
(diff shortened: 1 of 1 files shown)


thumphries:
The `**` refspec won't work until we upgrade the git version on boris :(

thumphries:
(it'll work for master and some branches, but won't traverse a `/`)

🔪  otherwise

jacobstanley:
yeah that's cool, I'm assuming that we will upgrade at some point and then it will work

PR #118, opened 2016-06-28
flock is hard, even (especially?) when wrapped in a fluffy API

jacobstanley:
Create the lock file before trying to lock on it.

The [filelock](https://hackage.haskell.org/package/filelock) package is designed in such a way that if the file doesn't exist, there will be a race and two separate fds can be created for the same filename. This means that both processes / threads successfully obtain the lock.

/cc @markhibberd @nhibberd @thumphries

### DIFF
--- a/src/Mafia/Install.hs
+++ b/src/Mafia/Install.hs
@@ -16,7 +16,7 @@ module Mafia.Install
   , renderInstallError
   ) where
 
-import           Control.Exception (SomeException)
+import           Control.Exception (SomeException, bracket)
 import           Control.Monad.IO.Class (MonadIO(..))
 import           Control.Parallel.Strategies (rpar, parMap)
 
@@ -54,7 +54,8 @@ import           P
 
 import           System.FileLock (SharedExclusive(..), FileLock)
 import           System.FileLock (lockFile, tryLockFile, unlockFile)
-import           System.IO (IO, stderr)
+import           System.IO (IO, FilePath, stderr)
+import           System.Posix.IO (OpenMode(..), OpenFileFlags(..), openFd, closeFd)
 
 import           Twine.Parallel (RunError(..), consume_)
 import           Twine.Data.Queue  (writeQueue)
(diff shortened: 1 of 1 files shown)


src/Mafia/Install.hs
```diff
@@ -446,13 +447,29 @@ lockPackage :: PackageEnv -> Package -> Flavour -> EitherT InstallError IO FileL
...
+      OpenFileFlags {
+          append = False
+        , exclusive = False
+        , noctty = False
+        , nonBlock = False
+        , trunc = False
+        }
+  in
+    bracket (openFd path ReadOnly (Just 0o666) flags) closeFd $ \_ ->
+      pure ()
```
erikd-ambiata, line 471: Use `const` maybe?

erikd-ambiata:
:+1: on :green_apple:

olorin:
Explanatory comment would be good, otherwise :+1:.

thumphries:
... that's quite bad, defeats the whole premise of FileLock. Anything we can do about it upstream?

jacobstanley:
> ... that's quite bad, defeats the whole premise of FileLock.

That's what I thought!

Yeah it should be fixable upstream, I think we just need to ensure the file exists before we get the fd to it.

Unix experts, does it make sense that two parallel calls to `open` can return different a fd?

olorin:
> Unix experts, does it make sense that two parallel calls to open can return different a fd?

Two parallel calls to open will return different fds. Not sure how it could be otherwise?

jacobstanley:
> Two parallel calls to open will return different fds. Not sure how it could be otherwise?

Sorry I don't mean fds, I mean inodes

olorin:
> Sorry I don't mean fds, I mean inodes

Makes sense to me - `open(3)[thread1]` then `unlink(3)`, `rename(3)` and then `open(3)[thread2]`, unless I'm misunderstanding.

jacobstanley:
No renaming or unlinking, I just mean both threads call `open(3)` at the same time with the same path and the file does not exist yet

markhibberd:
Mmm, yeh, I think my comment was wrong, there is something else subtle going on here that I am missing.

jacobstanley:
I could always switch to Max's [functor lock](https://github.com/ambiata/chum/blob/master/chum-ingest/main/Main.hs#L245-L271)

olorin:
> No renaming or unlinking, I just mean both threads call open(3) at the same time with the same path and the file does not exist yet

This can create two inodes if it's not opened with `O_EXCL` as well as `O_CREAT` - atomicity is guaranteed with `O_EXCL` but not otherwise afaik.

markhibberd:
Yeh, so that is one hole, the other is this pleasantly worded piece (http://linux.die.net/man/2/flock):

> An attempt to lock the file using one of these file descriptors may be denied by a lock that the calling process has already placed via another descriptor.

Hanging a lot on _may_. 

So horrid, complicated, but only real path might be touch + internal process lock as well.

jacobstanley:
Closing in favour of https://github.com/ambiata/mafia/pull/119

PR #119, opened 2016-06-28
flock is only process safe, you need to do your own locking for thread safety

jacobstanley:
Attempt number 2, this locks using a map of mvars before attempting a system wide lock.

/cc @markhibberd @olorin @thumphries @charleso

### DIFF
--- a/test/Test/Mafia/Flock.hs
+++ b/test/Test/Mafia/Flock.hs
@@ -1,0 +1,62 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell #-}
+module Test.Mafia.Flock where
+
+import           Control.Concurrent (threadDelay)
+import           Control.Concurrent.MVar (newEmptyMVar, readMVar, putMVar)
+import           Control.Concurrent.Async (async, mapConcurrently, wait)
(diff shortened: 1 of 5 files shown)


test/Test/Mafia/Flock.hs
```diff
@@ -0,0 +1,62 @@
...
+
+    x <- async $ mapConcurrently loop ([0..999] :: [Int])
+    threadDelay 50000 -- nick's fault
+    putMVar pin ()
+    _ <- wait x
+
+    n <- readIORef ref
+
+    pure $
+      n === 1000
```
jacobstanley, line 53: This test is pretty terrible, but it does fail consistently if you replace the locking with the identity function.

markhibberd, line 53: Interesting you, got it to fail on os-x and linux?

markhibberd:
🎱 

Probably deserves a :monster: as well, but yeh, unix hey. Never let someone tell you it is pretty.

test/Test/Mafia/Flock.hs
jacobstanley, line 53: As discussed IRL, I meant that the test fails with no locking at all, so the test actually does check something, despite being utterly terrible, but it's only really exercising the thread level locking.

olorin:
:+1:

nhibberd:
:penguin: :facepunch:

PR #121, opened 2016-06-29
Support for detecting and installing build-tools

jacobstanley:
Before trying to build packages which we are installing, mafia will now parse their `.cabal` file and install any build tools (i.e. `happy`, `alex`) which are required and also put them on the `PATH` temporarily.

This means we don't need `MAFIA_HAPPY=true` and friends anymore for the most part. It is still required for bmx itself as this only applies to dependencies, but we can fix easily enough in another PR I think.

It might be that we need some environment variable to control this magic and maybe exclude certain build tools, but I'm not sure what that looks like yet so I'd rather wait until we have a problem, and I'll fix it when it comes up.

/cc @markhibberd @thumphries @charleso @nhibberd @olorin

### DIFF
--- a/ambiata-mafia.cabal
+++ b/ambiata-mafia.cabal
@@ -22,6 +22,7 @@ library
                     , attoparsec                      >= 0.12       && < 0.14
                     , base16-bytestring               == 0.1.*
                     , bytestring                      == 0.10.*
+                    , Cabal                           >= 1.22       && < 1.25
                     , case-insensitive                == 1.2.*
                     , containers                      == 0.5.*
                     , cryptonite                      == 0.15.*
--- a/src/Mafia/Cabal/Dependencies.hs
+++ b/src/Mafia/Cabal/Dependencies.hs
@@ -182,7 +182,7 @@ makeInstallPlan mdir sourcePkgs installArgs = do
   withSystemTempDirectory "mafia-deps-" $ \tmp -> do
     let
       dir = fromMaybe tmp mdir
-      cabal = cabalFrom dir (Just (tmp </> "sandbox.config"))
+      cabal = cabalFrom dir (tmp </> "sandbox.config") []
 
     Hush <- cabal "sandbox" ["init", "--sandbox", tmp]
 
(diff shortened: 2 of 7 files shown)


ambiata-mafia.cabal
```diff
@@ -22,6 +22,7 @@ library
                     , attoparsec                      >= 0.12       && < 0.14
                     , base16-bytestring               == 0.1.*
                     , bytestring                      == 0.10.*
+                    , Cabal                           >= 1.22       && < 1.25
```
jacobstanley, line 25: :( the day has come

markhibberd, line 25: I can feel the CPP clouds forming

jacobstanley, line 25: 😨 If it comes to that I'll write a `.cabal` file parser 😟

erikd-ambiata, line 25: Sorry, what is the issue there? Why can't you use Cabal's cabal file parser?

jacobstanley, line 25: I just don't want to depend on the `Cabal` library if I can avoid it, it opens the door to mafia needing to be built again the same `Cabal` library as the `cabal-install` executable. Currently we can use the same mafia executable with different `cabal-install` versions and manage the differences dynamically.

jacobstanley, line 25: I think it shouldn't be too much of an issue in this case, because we're using a fairly stable part of the API here, but I'd like to keep it to a minimum.

thumphries:
~~It seems to use the transitive dependency hash when installing build tools, does that sound right? It should probably disregard the deps of wherever it's being built, or we'll end up with 65 copies of happy and alex.~~ ^ this was wrong, it just doesn't look at the `bin` symlinks we were using before. means we'll be using bleeding-edge happy and alex. they have a pretty small dependency footprint so it's not a big deal

You're right about build-tools being a list of package names. e.g. `hscolour` is the hackage name, `HsColour` is the binary name, and people do `build-tools: hscolour`, very convenient. I wonder if we can teach it to override cabal's burned-in list of build tools.

Code is 🍯

PR #122, opened 2016-06-29
Fix "cabal sandbox hc-pkg recache (exit code: 1)" when switching GHC versions

jacobstanley:
When switching between ghc versions, cabal-install 1.24 and later get confused about where the sandbox is supposed to be because they read the ghc version from `dist/setup-config` rather than looking at the ghc on the `PATH` to detect the version.

This fix dances around the issue and removes the `dist/setup-config` if it detects the problem, and this forces a `cabal configure` which populates `dist/setup-config` with the correct information.

/cc @thumphries @tmcgilchrist

### DIFF
--- a/src/Mafia/Init.hs
+++ b/src/Mafia/Init.hs
@@ -38,7 +38,7 @@ import           P
 
 import           System.IO (IO, stderr)
 
-import           X.Control.Monad.Trans.Either (EitherT, left, hoistEither)
+import           X.Control.Monad.Trans.Either (EitherT, runEitherT, left, hoistEither)
 
 
 data InitError =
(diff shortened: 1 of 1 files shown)


tmcgilchrist:
Running into this fairly regularly 🚀

erikd-ambiata:
:100: :100: :100: :100: :100: :100:

jacobstanley:
I have to say though, even with this implemented, it's annoying that it has to rebuild the entire project (all the `.o` files get waxed).

Might be worth investigating what it would take to keep separate `dist` directories, the same way we do sandboxes, so that you can pick up where you left off and keep all your `.o` files for all ghc versions.

jacobstanley:
We do have `dist/build/<project>/<project>` hard coded in a bunch of tests though, might not be worth the pain.

tmcgilchrist:
My general workflow is pick a version either 7.10 or 8 and do all dev in that, and if the build breaks switch to that version and check it out.

erikd-ambiata:
Isn't it a library problem rather than a final binary problem?

When using cabal sandboxes and switching compiler versions, all that gets build is the final linked binary. The libs are all in a compiler version specific tree like `.cabal-sandbox/lib/x86_64-linux-ghc-8.0.1`.

jacobstanley:
@erikd-ambiata yes when you're building libraries the modules get built to `dist/build/Foo/Bar.o` where the module name is `Foo.Bar`. If you switch between ghc versions, build, and then switch back, the module object file gets clobbered, would be nice if it didn't.

erikd-ambiata:
Ah right, thanks for the clarification.

PR #123, opened 2016-06-29
Put build-tools on path for current project

jacobstanley:
This is a follow on from the build-tools PR yesterday, doing the same thing for the current project.

It also removes the `MAFIA_HAPPY` like env vars are they are no longer required.

/cc @thumphries

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -398,7 +398,6 @@ mafiaHoogle args = do
 
 mafiaInstall :: InstallPackage -> EitherT MafiaError IO ()
 mafiaInstall ipkg = do
-  ensureAuxillaryExes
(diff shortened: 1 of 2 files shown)


tmcgilchrist:
Looks good 🐫

erikd-ambiata:
:goat:

thumphries:
🍔  that was fast!

Do we still need the `bin` symlinks?

jacobstanley:
> Do we still need the bin symlinks?

I tried getting rid of them, but the problem is that `ensureBuildTools` runs on every mafia command, and the symlinks are a fast way to check that we already have the build tool. It shaves about 2 seconds off the build.

Potentially we could put the symlinks in the sandbox instead, that might be a bit cleaner, but I'll leave that for another day.

PR #124, opened 2016-06-30
Support constraints on build-tools

jacobstanley:
Turns out that I need to support constraints after all, as `haskell-src-exts` has constraints on `happy`

/cc @thumphries


jacobstanley:
Arg how annoying, can't use jack in mafia because it needs happy lol

thumphries:
nice! 🐫

erikd-ambiata:
:goat: on :green_apple:

PR #132, opened 2016-08-24
Honour MAFIA_HOME environment variable in the shell script

jacobstanley:
/cc @nhibberd @thumphries @markhibberd @tmcgilchrist @olorin

### DIFF
--- a/script/mafia
+++ b/script/mafia
@@ -1,5 +1,7 @@
 #!/bin/sh -eu
 
+: ${MAFIA_HOME:=$HOME/.mafia}
+
 fetch_latest () {
   if [ -z ${MAFIA_TEST_MODE+x} ]; then
     TZ=$(date +"%T")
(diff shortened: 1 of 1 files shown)


script/mafia
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh -eu
 
+MAFIA_HOME=${MAFIA_HOME:-$HOME/.mafia}
+
```
jacobstanley, line 4: Not sure if there is a better way to do this?

erikd-ambiata: LGTM! It should work correctly with any POSIX shell.

markhibberd: "better" is relative, as @erikd-ambiata said, above will work perfectly. The only other one that you will see lots in scripts is:

```
: ${MAFIA_HOME:=$HOME/.mafia}
```

It isn't better/worse, just pointing it out because it occurs a fair bit.

jacobstanley: @markhibberd I think I like what you have better as it only mentions `MAFIA_HOME` once

Can you explain what it means? why have the `:` at the start?

markhibberd: `:` Is a shell builtin (`man builtins`) that is a command that always returns true, it is effectively a symbolic version of `true`, but by convention in most shell scripts (well BSD world anyway) to be read as "evaluate all of my arguments, ignore them and do nothing", the "evaluate all of my arguments" actually being just normal shell expansion/evaluation nothing really specific to `:`. After that it is just normal shell expansion, `${_:=_}` as in assign if not set, rather than `${_:-_}` as in return if not set. But you need the `:` otherwise the shell would run the output of `${}` as a command and you just want to ignore it.

jacobstanley: Perfect, thanks!

PR #133, opened 2016-08-25
Better error messages for 'cabal unpack' and 'cabal sdist' failures

jacobstanley:
We occasionally get errors due to `cabal unpack` or `cabal sdist` failing on the build bots, the existing errors weren't very useful as they did not include the output from cabal itself. This fixes that, now when `cabal unpack` fails you get something like:

```
Failed to unpack bifunctors-5.2.1 to /var/folders/ym/gt38kky12mng7gp0kvqqg5s40000gn/T/mafia-build-tools-72310
Process failed with exit code: 1
/foo: createDirectory: permission denied (Permission denied)
```

and when `cabal sdist` fails you get something like:

```
Failed to run 'cabal sdist' for source package: lib/p
Process failed with exit code: 1
cabal: Error: Could not find module: P.Foo.Bar.Baz with any suffix:
["gc","chs","hsc","x","y","ly","cpphs","hs","lhs"]
```

/cc @erikd-ambiata @nhibberd @charleso @olorin

### DIFF
--- a/src/Mafia/Cabal/Package.hs
+++ b/src/Mafia/Cabal/Package.hs
@@ -283,9 +283,11 @@ hashSourcePackage dir = do
 
   sdistFiles <- withSystemTempDirectory "mafia-sdist-" $ \tmp -> do
     let
-      sdistError _ = CabalSDistFailed rdir
       path = tmp </> "sources.txt"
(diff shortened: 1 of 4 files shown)


erikd-ambiata:
LGTM! :+1: on :green_apple:

nhibberd:
:100:

PR #136, opened 2016-08-29
Add --dump-core option

amosr:
This might not be the best way to go about it. Just an idea

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -216,7 +216,7 @@ pCoreDump :: Parser CoreDump
 pCoreDump =
   flag DisableCoreDump EnableCoreDump $
        long "dump-core"
-    <> help "Dump prepared Core output to dist/build/*."
+    <> help "Dump the optimised Core output to dist/build/*. This is simply a shorthand for other GHC options."
 
 pDependsUI :: Parser DependsUI
 pDependsUI =
@@ -359,8 +359,16 @@ mafiaBuild p w dump flags args = do
       case dump of
         DisableCoreDump ->
           []
-        EnableCoreDump ->
-          ["--ghc-options=-ddump-prep -ddump-to-file -dppr-case-as-let -dsuppress-all"]
+        EnableCoreDump -> fmap ("--ghc-options="<>)
+          ["-ddump-simpl"
+          ,"-ddump-to-file"
+          ,"-dppr-case-as-let"
+          ,"-dsuppress-uniques"
+          ,"-dsuppress-idinfo"
+          ,"-dsuppress-coercions"
+          ,"-dsuppress-type-applications"
+          ,"-dsuppress-module-prefixes"
+          ]
 
 
   liftCabal . cabal_ "build" $ ["-j"] <> wargs <> dumpargs <> args
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -71,7 +71,7 @@ data MafiaCommand =
   | MafiaHash
   | MafiaDepends DependsUI (Maybe PackageName) [Flag]
   | MafiaClean
-  | MafiaBuild Profiling Warnings [Flag] [Argument]
+  | MafiaBuild Profiling Warnings CoreDump [Flag] [Argument]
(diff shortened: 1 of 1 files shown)


amosr:
@jystic @charleso

jacobstanley:
this is cool, will take a proper look tomorrow

charleso:
I really know nothing about the flags in question.

Just curious, is it already possible to pass ghc flags though? I guess I'm asking - does this need to be in mafia or is it just a call-site decision?

HuwCampbell:
One can use arguments which are passed to cabal after the `--`, and `--ghc-options=*` will shuffle them there.

So I think this is for convenience as there's 4 flags required here.

jacobstanley:
Yeah it's definitely convenience, I for one can never remember what the flags are called.

I think maybe `-dsuppress-all` is a little too aggressive, I quite like having the type signatures in the output or I find it harder to decipher what's going on. Would you be ok with using `-dsuppress-coercions -dsuppress-idinfo -dsuppress-module-prefixes` in place of `-dsuppress-all`? It's basically the same but with type signatures.

What is the difference between between `-ddump-prep` and `-ddump-simpl`? For some reason `-ddump-prep` always gives me a file with two dumps right after another which appear to be identical?

I think I would prefer to use `-ddump-simpl` or maybe we should have two options in mafia `--dump-simpl` and `--dump-prep` (instead of `--dump-core`), what do you think? /not sure

jacobstanley:
> maybe we should have two options in mafia --dump-simpl and --dump-prep (instead of --dump-core), what do you think? /not sure

Having said this, I do like the simplicity of having only a single `--dump-core` option. Part of the benefit of this feature is that it makes it trivial to get started inspecting core, even if you've never done it before.

tranma:
I'm 👍 for `-ddump-core`

can also use something from this https://github.com/shachaf/ghc-core/blob/master/ghc-core.hs to make it more readable, if necessary

HuwCampbell:
It's similar to the profiling flag, in that there's scope in mafia for making things which ghc and cabal make hard easy.

main/mafia.hs
```diff
@@ -344,7 +355,15 @@ mafiaBuild p w flags args = do
         EnableWarnings ->
           ["--ghc-options=-Werror"]
 
-  liftCabal . cabal_ "build" $ ["-j"] <> wargs <> args
+    dumpargs =
+      case dump of
+        DisableCoreDump ->
+          []
+        EnableCoreDump ->
+          ["--ghc-options=-ddump-prep -ddump-to-file -dppr-case-as-let -dsuppress-all"]
```
jacobstanley: You can mention `--ghc-options` multiple times, so I think I would prefer this to be:

``` hs
[ "--ghc-options=-ddump-prep"
, "--ghc-options=-ddump-to-file"
, "--ghc-options=-dppr-case-as-let"
-- etc --
]
```

amosr: Yeah fair enough, I'm not too worried either way, will change

amosr:
Yeah, it really is just a shorthand. Maybe we want a more principled way of doing shortcuts that expand to multiple options?

amosr:
> What is the difference between between -ddump-prep and -ddump-simpl? For some reason -ddump-prep always gives me a file with two dumps right after another which appear to be identical?

It's just cargo-cult muscle memory for me. I think prep is slightly after simpl in the pipeline, but I don't think it makes a real difference. Happy to go with simpl

amosr:
> can also use something from this https://github.com/shachaf/ghc-core/blob/master/ghc-core.hs to make it more readable, if necessary

I just had a look, maybe those regexes are superseded by the -dsuppress options?

jacobstanley:
:wolf: so good

erikd-ambiata:
> I just had a look, maybe those regexes are superseded by the -dsuppress options?

@amosr That might be the case. If I remember correctly ghc-core has other features beyond just that. Its a pity its been un-maintained for so long.

PR #140, opened 2016-09-26
Add command to grab mafia script for the first time.

markhibberd:
For the lazy.

Don't actually care either way, just surprised it wasn't there
when I went looking. Merge if you think it is useful, close 
otherwise.


jacobstanley:
Sounds good to me 👍

PR #142, opened 2016-11-24
WIP gc

thumphries:
Really simple/stupid garbage collector POC for the mafia cache. Point it at your root directory (in my case `~/src` contains all my projects, I assume most people have something similar). it'll find any live symlinks to cached packages, record them as Used, and optionally print/delete all the unused packages in the cache. Lets you clean up your disk without interrupting your work (although, true, I stopped work for two hours to write this)

Usage: `mafia-gc ~/.ambiata/mafia ~/src`

Please ignore the PR structure. I have no intent to merge this in the current state, just wanted to share and get some feedback before I spend any time integrating it with mafia.

@olorin @nhibberd does this work on your box? (drop into ghci if you want it to delete stuff, I haven't written any optparse; xargs is fine too)

@jystic would this functionality be worth including in mafia, or is it too hacky / has too many dependencies?

It could easily be a shell script, but there are a lot of different versions of `find` out there, also I've not used `comm` before.

### DIFF
--- a/mafia-gc/Setup.hs
+++ b/mafia-gc/Setup.hs
@@ -1,0 +1,106 @@
+#!/usr/bin/env runhaskell
+
+import           Data.Char (isDigit, toLower)
+import           Data.Function (on)
+import           Data.List (intercalate, sortBy)
+import           Data.Monoid ((<>))
+import           Data.Version (showVersion)
+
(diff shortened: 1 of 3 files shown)


thumphries:
Seems this can't work properly until we change the way source dependencies are compiled

PR #143, opened 2016-12-30
Support for running Haskell "shell scripts"

jacobstanley:
This adds a new command `mafia script` which allows the use of mafia as a shebang interpreter, for example:

```hs
#!/usr/bin/env mafia script

{-# LANGUAGE NoImplicitPrelude #-}
{-# LANGUAGE OverloadedStrings #-}

{-# SUBMODULE ambiata/p@3ea83a82b058ba2e1dd216d9e7832fd49cf33dbd #-}

{-# PACKAGE ambiata-p #-}
{-# PACKAGE pretty-show #-}
{-# PACKAGE text #-}

import qualified Data.Text as T
import qualified Data.Text.Read as T

import           P

import           System.IO (IO, putStrLn)
import           System.Environment (getArgs)

import           Text.Show.Pretty (ppShow)

data Argument =
    Int !Int64
  | String !Text
    deriving (Show)

parseArgument :: Text -> Argument
parseArgument xs =
  case T.decimal xs of
    Left _ ->
      String xs
    Right (x, "") ->
      Int x
    Right _ ->
      String xs

main :: IO ()
main = do
  args <- fmap (parseArgument . T.pack) <$> getArgs
  putStrLn $ ppShow args
```

~~I'm not sure if `mafia shell` is the best name, as it's not really a shell, suggestions welcome.~~

I think this will be useful for sharing code snippets, and trying things out without having to start up a whole new project.

### DIFF
--- a/src/Mafia/Shell.hs
+++ b/src/Mafia/Shell.hs
@@ -1,0 +1,377 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+module Mafia.Shell (
+    ShellError(..)
+  , renderShellError
+  , runShell
+  ) where
+
+import           Control.Monad.IO.Class (MonadIO(..))
+
+import qualified Data.Attoparsec.Text as Atto
+import qualified Data.Map as Map
+import qualified Data.Text as T
+
+import           Mafia.Hash
+import           Mafia.Home
+import           Mafia.IO
+import           Mafia.Path
+import           Mafia.Process
+
+import           P
+
+import           System.IO (IO)
+
+import           X.Control.Monad.Trans.Either (EitherT, hoistMaybe, hoistEither)
+
+
+data Script =
+  Script {
+      scriptId :: !Text
+    , scriptHash :: !Hash
+    , scriptPath :: !File
+    , scriptPackages :: ![Package]
+    , scriptSubmodules :: ![Submodule]
+    , scriptMain :: !Text
+    } deriving (Eq, Ord, Show)
+
+newtype Package =
+  Package {
+      unPackage :: Text
+    } deriving (Eq, Ord, Show)
+
+data Submodule =
+  Submodule {
+      submoduleUser :: !Text
+    , submoduleRepo :: !Text
+    , submoduleCommit :: !(Maybe Text)
+    } deriving (Eq, Ord, Show)
+
+data Pragma =
+    PragmaPackage !Package
+  | PragmaSubmodule !Submodule
+    deriving (Eq, Ord, Show)
+
+data ShellError =
+    ShellProcessError !ProcessError
+  | ShellFileNotFound !File
+  | ShellGitParseError ![Char]
+    deriving (Show)
+
+renderShellError :: ShellError -> Text
+renderShellError = \case
+  ShellProcessError err ->
+    renderProcessError err
+
+  ShellFileNotFound path ->
+    "File not found: " <> path
+
+  ShellGitParseError err ->
+    "Failed parsing git output: " <> T.pack err
+
+pPragma :: Text -> Atto.Parser a -> Atto.Parser a
+pPragma pragma parser = do
+  _ <- Atto.string "{-#" *> Atto.skipSpace *> Atto.string pragma
+  x <- Atto.skipSpace *> parser <* Atto.skipSpace
+  _ <- Atto.string "#-}"
+  pure x
+
+pPackage :: Atto.Parser Pragma
+pPackage =
+  fmap PragmaPackage . pPragma "PACKAGE" $
+    Package . T.strip <$> Atto.takeWhile (/= '#')
+
+pAlphaNum :: Atto.Parser Text
+pAlphaNum =
+  Atto.takeWhile1 (Atto.inClass "a-zA-Z0-9")
+
+pHash :: Atto.Parser Text
+pHash =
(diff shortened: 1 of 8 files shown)


src/Mafia/Shell.hs
```diff
@@ -0,0 +1,373 @@
...
+    , scriptHash :: !Hash
+    , scriptPath :: !File
+    , scriptPackages :: ![Package]
+    , scriptSubmodules :: ![Submodule]
+    , scriptMain :: !Text
+    } deriving (Eq, Ord, Show)
+
+newtype Package =
+  Package {
+      unPackage :: Text
```
thumphries: Could this support version constraints?

src/Mafia/Shell.hs
```diff
@@ -0,0 +1,373 @@
...
+      mapMaybe takeSubmodule pragmas
+
+    sid =
+      renderHash $ hashText path
+  in
+    Script sid (hashText source) path packages submodules source
+
+renderSubmoduleUrl :: Submodule -> Text
+renderSubmoduleUrl s =
+  "git@github.com:" <> submoduleUser s <> "/" <> submoduleRepo s
```
thumphries: Burning github in?

Any thoughts on supporting a full git remote url + refspec?

src/Mafia/Shell.hs
thumphries: I'm concerned this will lead to some fairly brittle scripts if it just jumps to the latest version each time. Would make it unsafe to use in prod (although I wouldn't be doing that, maybe it isn't a problem)

thumphries:
I like it. Dependency stuff seems a little sketchy if you're going to use it in prod.

Name suggestions, `mafia run` `mafia script` `mafia exec`

src/Mafia/Shell.hs
jacobstanley: This actually does support constraints, I'm just not parsing them, the string you put in the `PACKAGE` pragma is transplanted directly in to a `.cabal` file. I've updated the cli test to demonstrate.

I don't think they are safe to use in prod anyway because they'd depend on hackage and github being up.

src/Mafia/Shell.hs
jacobstanley: > Burning github in?

I thought it should at least be available as a shorthand

> Any thoughts on supporting a full git remote url

Something we can possibly add at a later date without breaking anything? (I have no use for it personally at the moment)

> refspec

I'm not sure how to quickly verify that the refspec which is checked out is the one we want.

With full hashes it's easy to compare to the output of `git submodule`, and with no refspec I assume it's always right as long as the submodule exists.

If you said say `@topic/foo` and the checked out submodule was in fact equivalent to the remote `topic/foo` that could be enough, but maybe the remote has been updated so you really ought to update the remote info, which is too time consuming to do every execution. I couldn't think of a robust solution with the right semantics so in the end I decided to just start with git hashes or nothing at all.

jacobstanley:
> I like it. Dependency stuff seems a little sketchy if you're going to use it in prod.

I can't how we can use it in prod at this stage, requiring hackage/github access seems like a deal breaker.

src/Mafia/Shell.hs
thumphries: good point

src/Mafia/Shell.hs
thumphries: I'll probably only ever use it for github remotes, but would prefer not to bake github into the tool. `git@github.com/` isn't a lot of typing. Not a strongly held opinion, just wondering what others think.

> I'm not sure how to quickly verify that the refspec which is checked out is the one we want.

fair enough. hashes are good

jacobstanley: > I'll probably only ever use it for github remotes, but would prefer not to bake github into the tool. `git@github.com/` isn't a lot of typing. Not a strongly held opinion, just wondering what others think.

I thought all the cool kids were baking github in these days 😂 

I'll see what I can cook up, I'm relying on only supporting github at the moment for unique paths on disk, the format is `lib/<user>/<repo>` and from that I can reverse engineer the required url for the submodule, but I can come up with a different scheme.

thumphries: Ah, that sounds like *work*. I won't actually need this feature, so ignore the heck out of me if you like. Thought it was just a matter of strings.

Also finally understand what you meant about adding it later without breakage (saw the same feature in `bower` and it seems to work, although it assumes `https` over `git`)

amosr:
yes please 👍

amosr:
yeah I think it's worth being very explicit that this isn't supposed to be for production (which were my only qualms originally)

tranma:
I like how agile this is. mafia agile. 🔫

src/Mafia/Shell.hs
jacobstanley: > saw the same feature in `bower` and it seems to work, although it assumes `https` over `git`

I assumed ssh so that it will work nicely with our private repos, not sure though

jacobstanley:
> Name suggestions, `mafia run` `mafia script` `mafia exec`

Decided on `mafia script`, I think `mafia run` should probably do what `cabal run` does if it were to exist

domdere:
holy crap this is awesome

jacobstanley:
I just realised this doesn't work on [Linux](http://unix.stackexchange.com/questions/63979/shebang-line-with-usr-bin-env-command-argument-fails-on-linux) because `env` can only take a single argument, e.g. `/usr/bin/env mafia` instead of `/usr/bin/env mafia script`, fix incoming.

PR #144, opened 2017-01-02
'mafia script' now works on Linux

jacobstanley:
I had to re-jig this so that we take the path to the script as the first argument as `/usr/bin/env` on Linux only takes a single argument.

A script path is detected using the following logic:
```hs
-- | Detect if an argument is a path to a script.
--
--   The most robust way I can think of is to look for a slash:
--
--   - Sub-commands will never contain a slash.
--
--   - If the script is on the PATH we'll receive the absolute path which will
--     contain a slash.
--
--   - If it's not on the PATH then the user will need to include a slash in
--     the relative path in order to execute it anyway, e.g. ./script or dir/script
--
isScriptPath :: FilePath -> Bool
isScriptPath =
  List.isInfixOf "/"
```

And if we are passed a command that we don't understand and which is not a script path then we produce the following error:
```
$ mafia myscript.hs
Invalid argument 'myscript.hs', not a valid command or a valid script path.
Note: paths to scripts must contain a slash, e.g. ./myscript.hs or /usr/bin/myscript.hs

Usage: mafia ((-v|--version) | (COMMAND | SCRIPT_PATH [SCRIPT_ARGUMENTS])
```

! @amosr @thumphries @tranma


main/mafia.hs
```diff
@@ -144,7 +145,12 @@ run = \case
     mafiaScript path args
 
 parser :: Parser MafiaCommand
-parser = subparser . mconcat $ commands
+parser =
+  subparser (mconcat commands) <|> pScript
+
+pScript :: Parser MafiaCommand
+pScript =
+  MafiaScript <$> pScriptPath <*> many pScriptArgs
```
amosr, line 157: confused - is this ever triggered? seems like this case is caught earlier by line 60

amosr: oh - it's only for help text?

jacobstanley: yeah, that's right, just for the help text, although there's an optparse PR up at the moment which will allow us to avoid the hack at the top (see link in the code)

amosr:
yep groovy 👍 

not a fan of ambiguous command lines that can take a filename or a command name, so I say requiring the slash is the best choice.

main/mafia.hs
```diff
@@ -304,15 +311,41 @@ pConstraint =
...
   argument textRead $
        metavar "SCRIPT_ARGUMENTS"
-    <> help "Argument to pass to the shell script."
+    <> help "Arguments to pass to the script."
+
+scriptRead :: ReadM File
+scriptRead =
+  eitherReader $ \path ->
+    if isScriptPath path then
+      Right $ T.pack path
```
HuwCampbell: This can't be triggered at this stage right? Should it be `readerError` to make that clear?

HuwCampbell: I guess the comment already makes that clear, so maybe not.

jacobstanley: Hmm, it's true, although I'm hoping it will be become the real code path in optparse 0.14.

I think you're right though, having dead code isn't a good idea. Curious, how does `readerError` differ from `Left` in this situation?

HuwCampbell: It's identical

```haskell
eitherReader f = readerAsk >>= either readerError return . f
```

main/mafia.hs
```diff
@@ -304,15 +311,41 @@ pConstraint =
...
+--
+--   - Sub-commands will never contain a slash.
+--
+--   - If the script is on the PATH we'll receive the absolute path which will
+--     contain a slash.
+--
+--   - If it's not on the PATH then the user will need to include a slash in
+--     the relative path in order to execute it anyway, e.g. ./script or dir/script
+--
+isScriptPath :: FilePath -> Bool
```
thumphries, line 349: 👍

main/mafia.hs
```diff
@@ -304,15 +311,44 @@ pConstraint =
...
+    <> help "Arguments to pass to the script."
+
+scriptRead :: ReadM File
+scriptRead =
+  eitherReader $ \path ->
+    if isScriptPath path then
+      Left $
+        "Something went wrong, '" <> path <> "' looks like a script, but script" <>
+        " execution should have been handled in an earlier code path.\n" <>
+        "Please report this as a bug: https://github.com/ambiata/mafia/issues"
```
jacobstanley, line 331: @HuwCampbell I've added an error message here instead, looks like this:
```
Something went wrong, './foo.hs' looks like a script, but script execution should have been handled in an earlier code path.
Please report this as a bug: https://github.com/ambiata/mafia/issues
```

Issue #35, opened 2016-01-04
quick doesn't support CPP version macros


charleso:
```
Foo.hs:15:21:
     error: token is not a valid binary operator in a preprocessor subexpression
if MIN_VERSION_time(1,5,0)
    ~~~~~~~~~~~~~~~~^
1 error generated.
```

jacobstanley:
some relevant info http://stackoverflow.com/questions/12546757/how-do-i-use-cabals-min-version-and-other-macros-with-ghci/12546758#12546758

erikd-ambiata:
@charleso Pretty sure that works out of the box with ghc-8.0 but not with ghc-7.10 and earlier.

jacobstanley:
@erikd-ambiata 👍 just checked this out, fails with 7.10, works with 8.0

PR #145, opened 2017-01-03
Include cabal_macros.h when launching 'mafia quick'

jacobstanley:
This allows usages of `CPP` which rely on version macros in GHC 7.10 and earlier, when loading a file with `quick`.

GHC 8.0 already does something similar by itself.

Fixes https://github.com/ambiata/mafia/issues/35

! @nhibberd @charleso

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -504,7 +504,7 @@ ghciArgs extraIncludes paths = do
 
   let
     dirs =
-      ["src", "test", "gen", "dist/build/autogen"] <> extras
+      ["src", "test", "gen", "dist/build", "dist/build/autogen"] <> extras
 
   headers <- getHeaders
   includes <- catMaybes <$> mapM ensureDirectory dirs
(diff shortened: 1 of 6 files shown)


charleso:
Fuck. Yes.

💯

charleso:
🎂

jacobstanley:
Ok, so this will now do nothing at all from `8.0.1` onwards, but for older versions it will load `cabal_macros.h`, assuming it exists. It's not the ideal fix, but probably good enough.

Note that in 7.10 you will need to run `mafia build` at least once before `cabal_macros.h` appears in `dist`.

main/mafia.hs
```diff
@@ -500,14 +502,39 @@ ghciArgs extraIncludes paths = do
 
   extras <- concat <$> mapM reifyInclude extraIncludes
 
-  let dirs = ["src", "test", "gen", "dist/build/autogen"] <> extras
-  includes  <- catMaybes <$> mapM ensureDirectory dirs
+  let
+    dirs =
+      ["src", "test", "gen", "dist/build/autogen"] <> extras
```
thumphries: any chance of adding`dist/build` while you're in here? Generated modules via Happy etc show up there and cause problems in `quick`

jacobstanley: Yeah I have the same issue with `hsc2hs`, it would be `dist/build` recursively right?

thumphries: I *think* you still need the extra one for `dist/build/autogen` (load path is dumb) but will check

thumphries: definitely need both

jacobstanley: oh, right no I meant do I need to include all the sub-directories of `dist/build` recursively, but no, only `dist/build` and `dist/build/autogen`, cool

thumphries: As per hipchat GHCi doesn't know how to rebuild these generated modules, so it won't reload the way you might expect. however right now it never works and transitively poisons all user code, so this is a far lesser evil

PR #147, opened 2017-01-11
test: Use HTTPS instead of GIT protocol

erikd-ambiata:
Mainly so that people outside Ambiata can build/run the tests.

Signed-off-by: Erik de Castro Lopo <erik.decastrolopo@ambiata.com>

! @jystic

### DIFF
--- a/test/cli/basic-usage/run
+++ b/test/cli/basic-usage/run
@@ -22,7 +22,7 @@ check() {
 
   cd ${MAFIA_TEMP}
   if [ ! -d "${MAFIA_REPO}" ]; then
-    git clone git@github.com:ambiata/${MAFIA_REPO}
+    git clone https://github.com/ambiata/${MAFIA_REPO}
   fi
 
   cd ${MAFIA_DIR}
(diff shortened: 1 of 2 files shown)


markhibberd:
You don't need ambiata creds to use ssh, only an ssh key known to GitHub. So people can build and run tests outside as it is, admittedly https doesn't require any creds but makes it significantly more painful for people who have write access and use ssh so I am slightly skeptical. At some point submodules are always going to be easier for the people who use the same proptocol as the submodule by default and more difficult for everyone else. But it is still fairly straight forward to do by either adding a key to use ssh or by using git config to switch submodule protocol. Would a readme note explaining how to do it achieve the same thing, without making it harder for people who prefer ssh access?

erikd-ambiata:
@markhibberd I agree with what you say in general, but in this specific case the changes I'm making are in the cli test scripts which do for instance:
```
git clone https://github.com/ambiata/p.git
```
Since this is only a test, that cloned repo is going to be used by the test and then deleted. That means its pretty much irrelevant whether it uses the Git or HTTPS protocol.

markhibberd:
Yep, fair enough.

jacobstanley:
👍 after rebase

jacobstanley:
@erikd-ambiata I'm merging this ahead of my other changes

PR #149, opened 2017-01-11
Add rebase check and fix for new whitelist

jacobstanley:
Will need to merge https://github.com/ambiata/mafia/pull/148 first, but after that this should be good.

! @erikd-ambiata @charleso


charleso:
👍

PR #151, opened 2017-01-16
Topic/exec

erikd-ambiata:
! @jystic

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -89,6 +89,7 @@ data MafiaCommand =
   | MafiaHoogle [Argument]
   | MafiaInstall [Constraint] InstallPackage
   | MafiaScript Path [Argument]
+  | MafiaExec [Argument]
     deriving (Eq, Show)
 
 data Warnings =
@@ -145,6 +146,8 @@ run = \case
     mafiaInstall ipkg constraints
   MafiaScript path args ->
     mafiaScript path args
+  MafiaExec args ->
+    mafiaExec args
 
 parser :: Parser MafiaCommand
 parser =
@@ -216,6 +219,10 @@ commands =
  , command' "install" ( "Install a hackage package and print the path to its bin directory. "
                      <> "The general usage is as follows:  $(mafia install pretty-show)/ppsh" )
             (MafiaInstall <$> many pConstraint <*> pInstallPackage)
+
+ , command' "exec" ( ghciText <> " Exec the provided command line in the local cabal sandbox." )
+            (MafiaExec <$> many pCabalArgs)
+
  ]
   where
     ghciText = "Start the repl directly skipping cabal, this is useful "
@@ -501,6 +508,18 @@ mafiaInstall ipkg constraints = do
 mafiaScript :: File -> [Argument] -> EitherT MafiaError IO ()
 mafiaScript file args =
   firstT MafiaScriptError $ runScript file args
+
+mafiaExec :: [Argument] -> EitherT MafiaError IO ()
+mafiaExec args = do
+  let fixedArgs =
+        case args of
+          [] -> [] -- Should not happen.
+          [x] -> [x] -- A command without arguments.
+          -- Insert `--` between the command and its arguments so cabal doesn't
+          -- mess with them.
+          (x:xs) -> x : "--" : xs
+  exec MafiaProcessError "cabal" $ "exec" : fixedArgs
+
 
 ghciArgs :: [GhciInclude] -> [File] -> EitherT MafiaError IO [Argument]
 ghciArgs extraIncludes paths = do
--- a/test/cli/exec/run
+++ b/test/cli/exec/run
@@ -1,0 +1,31 @@
+#!/bin/sh -eux
+
+absolute_path() {
+  echo "$(cd "$(dirname "$1")" && pwd)/$(basename "$1")"
+}
+
+MAFIA=$(absolute_path ${1:-./dist/build/mafia/mafia})
+
+MAFIA_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'mafia')
+trap "rm -rf \"${MAFIA_TEMP}\"" EXIT
+
+INPUT=$(dirname $(absolute_path $0))
+
+echo "INPUT $INPUT"
+
+cd ${MAFIA_TEMP}
+
+cp ${INPUT}/trivial/* .
+
+# Mafia wants to be in a current git tree, so create one.
+git init
+git add *
+git commit -m "Initial commit"
+
+# Sets up sandbox, nstalls packages and builds project.
+mafia build
+
+# Create a new Haskell file not in the cabal file so we can build it.
+cp trivial.hs xxx.hs
+
+MAFIA_HOME=${MAFIA_TEMP} ${MAFIA} exec -- ghc -Wall xxx.hs -o xxx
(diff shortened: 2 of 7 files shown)


charleso:
No opinion, I've never needed this (that I know of). Deferring to Jacob.

main/mafia.hs
```diff
@@ -502,6 +509,17 @@ mafiaScript :: File -> [Argument] -> EitherT MafiaError IO ()
...
+mafiaExec :: [Argument] -> EitherT MafiaError IO ()
+mafiaExec args =
+  liftCabal $ cabal_ "exec" $ fixupArgs args
+  where
+    fixupArgs = \case
+      [] -> [] -- Should not happen.
+      [x] -> [x] -- A command without arguments.
+      -- Insert `--` between the command and its arguments so cabal doesn't
+      -- mess with them.
+      (x:xs) -> x : "--" : xs
```
jacobstanley: I'm good with this, just a minor nit, I'm trying to get rid of uses of `where` from this codebase, so if you don't mind can we change it to something like:
```hs
mafiaExec :: [Argument] -> EitherT MafiaError IO ()
mafiaExec args =
  let
    fixupArgs = \case
      [] ->
        [] -- Should not happen.
      [x] ->
        [x] -- A command without arguments.
      -- Insert `--` between the command and its arguments so cabal doesn't
      -- mess with them.
      x : xs ->
        x : "--" : xs
  in
    liftCabal $ cabal_ "exec" $ fixupArgs args
```

jacobstanley:
Can you add a cli test for this? There should be one for `mafia hoogle` that you can copy

thumphries:
Feature seems reasonable. I've used `cabal exec` in a test suite before, where I've relied on previous `mafia build` commands performing the mafia init for me.

main/mafia.hs
amosr: why is that?

erikd-ambiata: To run `sh -c 'echo Hello'` with cabal you would do:
```
cabal exec sh -- -c 'echo Hello'
```
so that cabal knows not to process command line args after `--`. If I don't insert `--`, cabal will treat the `-c` as one of the cabal arguments.

amosr: oops: I meant why is @jystic trying to get rid of `where`s?

tranma: he's a Standard ML missionary now

erikd-ambiata:
Test added.

olorin:
:cat:

main/mafia.hs
jacobstanley: > why is that?

I think it makes a function easier to read if you can follow it from top to bottom rather jumping around to follow the flow of data.

It is also more consistent with many of our other projects. When I first started at Ambiata I used `where` a lot, but after reading the source of a bunch of our projects I found that it was not used at all and the resulting code was easier to follow as a result, so I decided to stop using it myself in the name of consistency.

test/cli/exec/run
```diff
@@ -0,0 +1,14 @@
...
+}
+
+MAFIA=$(absolute_path ${1:-./dist/build/mafia/mafia})
+
+MAFIA_TEMP=$(mktemp -d 2>/dev/null || mktemp -d -t 'mafia')
+trap "rm -rf \"${MAFIA_TEMP}\"" EXIT
+
+cd ${MAFIA_TEMP}
+
+MAFIA_HOME=${MAFIA_TEMP} ${MAFIA} exec -- ghc-pkg list base
```
jacobstanley: How do you know this is working? `ghc-pkg list base` would also succeed if you ran it from outside a sandbox

erikd-ambiata: That test is testing the `exec` rather than the sandbox. If there's *anything* wrong with the `exec` it will fail. I'm trusting cabal to do the right thing (maybe thats not wise).

erikd-ambiata: Testing more than just the `exec` would require a different approach. Certainly possible though.

main/mafia.hs
```diff
@@ -502,6 +509,18 @@ mafiaScript :: File -> [Argument] -> EitherT MafiaError IO ()
...
+mafiaExec :: [Argument] -> EitherT MafiaError IO ()
+mafiaExec args = do
+  let fixedArgs =
+        case args of
+          [] -> [] -- Should not happen.
+          [x] -> [x] -- A command without arguments.
+          -- Insert `--` between the command and its arguments so cabal doesn't
+          -- mess with them.
+          (x:xs) -> x : "--" : xs
+  liftCabal $ cabal_ "exec" fixedArgs
```
jacobstanley: Sorry, I missed this the first time, you probably want to use `Mafia.Process.exec` here rather than `cabal_`, so that we exec the `cabal` executable rather than just calling out to it.

test/cli/exec/run
jacobstanley: Hmm, yeah, I don't mind if you just want to keep it like this, would be nice to actually check it's doing what we want, but I see your point that it's just testing we plumb the right arguments through to cabal.

jacobstanley:
Looks like the 7.8 tests are failing because `x-eithert` no longer builds on 7.8

I wonder if we should drop support for using mafia with 7.8

erikd-ambiata:
I thought everything built with 7.10 now. In fact most stuff builds with 8.0.

main/mafia.hs
erikd-ambiata: Fixed.

test/cli/exec/run
erikd-ambiata: Ok, I now have a much more comprehensive test (that runs in under 5 seconds).

erikd-ambiata:
As for `x-eithert` we either build that with 7.8 or we drop 7.8 support for mafia.

jacobstanley:
Yeah, so it's nice knowing that mafia will work with 7.8 so it can be used for any old random project.

It's easy to fix `x-eithert` so it still builds on 7.8, we just need to use `mapM` instead of `traverse`. But arguably the test that verifies mafia with the latest `p` and `x` is not sensible, maybe it should be for a specific commit, which could be different depending on the ghc version.

test/cli/exec/trivial/trivial.cabal
```diff
@@ -0,0 +1,16 @@
...
+license-file:        LICENSE
+author:              Ambiata
+maintainer:          info@ambiata.com
+build-type:          Simple
+cabal-version:       >=1.10
+
+executable trivial
+  default-language:    Haskell2010
+  main-is:             trivial.hs
+  build-depends:       base >=4.9 && <4.10
```
jacobstanley: Looks like these bounds need to be wider for 7.8 and 7.10

erikd-ambiata: Yes, I'll fix that!

test/cli/exec/run
```diff
@@ -0,0 +1,31 @@
...
+git add *
+git commit -m "Initial commit"
+
+# Sets up sandbox, nstalls packages and builds project.
+mafia build
+
+# Create a new Haskell file not in the cabal file so we can build it.
+cp trivial.hs xxx.hs
+
+MAFIA_HOME=${MAFIA_TEMP} ${MAFIA} exec -- ghc -Wall xxx.hs -o xxx
```
jacobstanley, line 31: This is much better

jacobstanley:
👍 on 🍏

PR #152, opened 2017-01-19
Add -p (--project) flag to mafia ghci/quick/watch

jacobstanley:
This flag automatically adds all of the current project's (i.e. git repository) source directories to the ghci include path.

So if I'm in project `foo`, in the `foo-cli` directory:
```
foo-data
└─ src
   └─ Foo
      └─ Data.hs
foo-cli
├─ main
│  └─ foo.hs
└─ src
   └─ Foo
      └─ Commands.hs
```
and I run:
```
~/src/foo/foo-cli $ mafia ghci -p main/foo.hs
```
Then not only will I get `foo-cli/src` loaded in the interpreter, I will also get `foo-data/src` loaded in the interpreter.

This is similar to the `-a` / `--all` flag which already exists, but that loads all of submodules in the interpreter as well, which is often overkill and sometimes doesn't work for various reasons.

! @charleso @erikd-ambiata @thumphries @nhibberd

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -104,6 +104,7 @@ data CoreDump =
 
 data GhciInclude =
     Directory Directory
+  | ProjectLibraries
   | AllLibraries
     deriving (Eq, Show)
 
@@ -281,14 +282,26 @@ pGhciEntryPoint =
     <> help "The entry point for GHCi."
 
 pGhciIncludes :: Parser [GhciInclude]
-pGhciIncludes = many (pGhciIncludeDirectory <|> pGhciIncludeAllLibraries)
+pGhciIncludes =
+  fmap concat $ sequenceA [
+      many pGhciIncludeDirectory
+    , toList <$> optional pGhciIncludeProjectLibraries
+    , toList <$> optional pGhciIncludeAllLibraries
+    ]
+
+pGhciIncludeProjectLibraries :: Parser GhciInclude
+pGhciIncludeProjectLibraries =
+  flag' ProjectLibraries $
+       long "project"
+    <> short 'p'
+    <> help "Make all project source directories available for GHCi, does not include submodules."
 
 pGhciIncludeAllLibraries :: Parser GhciInclude
 pGhciIncludeAllLibraries =
   flag' AllLibraries $
        long "all"
     <> short 'a'
-    <> help "Include all library source directories for GHCi."
+    <> help "Make all source directories available for GHCi, even from submodules."
 
 pGhciIncludeDirectory :: Parser GhciInclude
 pGhciIncludeDirectory =
@@ -529,7 +542,7 @@ ghciArgs extraIncludes paths = do
 
   let
     dirs =
-      ["src", "test", "gen", "dist/build", "dist/build/autogen"] <> extras
+      standardSourceDirs <> extras
 
   headers <- getHeaders
   includes <- catMaybes <$> mapM ensureDirectory dirs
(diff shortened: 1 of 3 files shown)


erikd-ambiata:
Seems reasonable. Not sure i'd use it because I've never felt the need to do it manually.

:+1:

jacobstanley:
> Seems reasonable. Not sure i'd use it because I've never felt the need to do it manually.

It's really only useful if you're working in projects that a split in to multiple cabal libraries, like `mismi`, `icicle`, `amp`, `blizzard`, etc.

It means you can change code in say `mismi-core` while you're deving `mismi-s3` and you can reload with `:r` rather than having to rebuild `mismi-core` again.

erikd-ambiata:
> It's really only useful if you're working in projects that a split in to multiple cabal libraries, like mismi, icicle, amp, blizzard, etc.

And Walrus, Blowfish, Goat, Cleaver, Gettysburg , ...... :)

main/mafia.hs
```diff
@@ -281,7 +282,15 @@ pGhciEntryPoint =
     <> help "The entry point for GHCi."
 
 pGhciIncludes :: Parser [GhciInclude]
-pGhciIncludes = many (pGhciIncludeDirectory <|> pGhciIncludeAllLibraries)
+pGhciIncludes =
+  many (pGhciIncludeDirectory <|> pGhciIncludeProjectLibraries <|> pGhciIncludeAllLibraries)
```
charleso: Seems weird that you can say `-p -p`? Does this version below make sense?

```haskell
map join . sequence $ [
    many pGhciIncludeDirectory
  , pure <$> pGhciIncludeProjectLibraries
  , pure <$> pGhciIncludeAllLibraries
  ]
```

jacobstanley: yeah I think so, that's always bugged me tbh

jacobstanley: hmm that won't work as there is no `Monad` instance, not sure if something similar is possible though

jacobstanley:
> And Walrus, Blowfish, Goat, Cleaver, Gettysburg , ...... :)

Maybe you will find it useful then :)

thumphries:
Wouldn't use this for quick, but for `watch` it looks pretty useful, nice!

jacobstanley:
Curious how people use quick such that this wouldn't be useful?

The reason it's useful for me is that in my `init.vim` I have:
```vim
autocmd FileType haskell nnoremap <silent> <LocalLeader>q :T :q<CR>:T mafia quick -p "%"<CR>
```

The means that when I hit `;q` it quits the current ghci session and restarts quick on the file in the current buffer.

markhibberd:
:100: I currently have a bunch of alias' over quick that could be drastically simplified or removed with this.

nhibberd:
Best :+1: use multiple `-i`s a lot

main/mafia.hs
```diff
@@ -281,14 +282,26 @@ pGhciEntryPoint =
     <> help "The entry point for GHCi."
 
 pGhciIncludes :: Parser [GhciInclude]
-pGhciIncludes = many (pGhciIncludeDirectory <|> pGhciIncludeAllLibraries)
+pGhciIncludes =
+  fmap concat $ sequenceA [
+      many pGhciIncludeDirectory
+    , maybe [] (:[]) <$> optional pGhciIncludeProjectLibraries
```
charleso: `maybeToList`?

jacobstanley: can even just be `toList` I think

PR #153, opened 2017-01-20
Remove inconsistent usages of project

jacobstanley:
Previously mafia used the word project and package somewhat interchangeably. I want to use the word project to refer to a group of packages within in a single repository as per https://github.com/ambiata/mafia/pull/152. This is also the same terminology used by cabal starting with 1.24 which supports `cabal.project` files that reference multiple packages.

`getProjectName` and the whole `Mafia.Project` module has been removed in favour of `getCabalFile` which doesn't basically the same thing.

! @charleso @erikd-ambiata @thumphries


thumphries:
👍 
/jury approve

erikd-ambiata:
:100: for consistency!

PR #157, opened 2017-02-08
Support for running quick/ghci/watch from anywhere

jacobstanley:
This change makes it possible to run `mafia quick` (and friends) from inside any working directory and have it automatically use the right sandbox.

e.g.
```
$ ls
icicle-compiler     icicle-core     icicle-data     icicle-repl     icicle-source
$ mafia quick icicle-compiler/src/Icicle/Compiler.hs
*Icicle.Compiler
λ
```

Note that the `icicle-compiler` sandbox was automatically selected without having to change to that directory.

This is particular useful when combined with an editor which can involve the repl on the currently active buffer.

/jury approved @tranma @nhibberd

### DIFF
--- a/src/Mafia/Cabal/Package.hs
+++ b/src/Mafia/Cabal/Package.hs
@@ -11,6 +11,7 @@ module Mafia.Cabal.Package (
   , getPackageId
   , getPackageType
   , getCabalFile
+  , findCabalRoot
 
   , SourcePackage(..)
   , getSourcePackage
@@ -69,6 +70,27 @@ getCabalFile dir = do
       return x
     _ : _ ->
       left $ CabalMultipleFilesFound dir xs
+
+-- | Given a file or directory path, try to find the .cabal file by searching
+--   its parent directories.
+findCabalRoot :: MonadIO m => Path -> m (Maybe Directory)
+findCabalRoot path0 = do
+  path <- canonicalizePath path0
+  entries <- tryGetDirectoryContents path
+  if any (\x -> extension ".cabal" x && x /= ".cabal" && not (T.isSuffixOf "-test.cabal" x)) entries then
+    return $ Just path
+  else if path == "/" then
+    return Nothing
+  else
+    findCabalRoot $ takeDirectory path
+
+tryGetDirectoryContents :: MonadIO m => Path -> m [File]
+tryGetDirectoryContents path = do
+  isDir <- doesDirectoryExist path
+  if isDir then
+    getDirectoryContents path
+  else
+    return []
 
 isCabalFile :: MonadIO m => File -> m Bool
 isCabalFile file =
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -506,15 +506,17 @@ mafiaUnlock = do
 
 mafiaQuick :: [Flag] -> [GhciInclude] -> [File] -> EitherT MafiaError IO ()
 mafiaQuick flags extraIncludes paths = do
-  args <- ghciArgs extraIncludes paths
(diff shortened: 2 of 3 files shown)


nhibberd:
Sounds legit :+1:

src/Mafia/Cabal/Package.hs
```diff
@@ -67,6 +68,27 @@ getCabalFile dir = do
     _ : _ ->
       left $ CabalMultipleFilesFound dir xs
 
+-- | Given a file or directory path, try to find the .cabal file by searching
+--   its parent directories.
+findCabalRoot :: MonadIO m => Path -> m (Maybe Directory)
+findCabalRoot path0 = do
+  path <- canonicalizePath path0
+  entries <- tryGetDirectoryContents path
+  if any (\x -> extension ".cabal" x && x /= ".cabal" && not (T.isSuffixOf "-test.cabal" x)) entries then
```
thumphries, line 80: Feels like you're burning in an ambiata convention here

thumphries:
Not a fan of this if it takes priority over the sandbox in `pwd`, feels weird. feature creep.

makes a lot of sense as a fallback when you're not in a cabal project though.

~~n.b. could you maybe write a shell script/vimscript to figure out the nearest cabal file and set that as CWD before running mafia? that's what we did in mafia-mode/stole from intero.~~ literally what you have in this PR lol

tranma:
👍 same as expected behaviour from other tools

src/Mafia/Cabal/Package.hs
jacobstanley: yeah I really am :/ although this is not new for quick/ghci/watch, we already bake in our standard source dir locations:
```hs
standardSourceDirs :: [Directory]
standardSourceDirs =
  ["src", "test", "gen", "dist/build", "dist/build/autogen"]
```

jacobstanley:
> Not a fan of this if it takes priority over the sandbox in `pwd`, feels weird. feature creep.

It would be a change in behaviour, but it would be consistent in that with this PR:
 - If you specify a path in a command it will find the closest parent `.cabal` file to the path
 - If you don't specify a path in a command it operates in the current working dir

Is this going to break your ghci script?

thumphries:
Nothing will break for me, since you've got that exception for `test.cabal` burned in. The action at a distance feels quite unusual, given other Haskell tools always prioritise the local sandbox if it exists. This makes most sense to me as a way to proceed when there's no cabal file or sandbox around.

Reaching into local subprojects is already pretty easy with `-p` and `-a`. I guess I don't know when you'd want to reach into and resandbox a totally foreign project when you're already in one, and if you did want to do that, you could `cd` or be explicit about it with a `quick -C` chdir flag.

Your codebase though, just registering my dissent :)

src/Mafia/Cabal/Package.hs
thumphries: Could we get src and test from the cabal file? or would that slow it down too much

Hit a similar problem with hinty yesterday, burning in our search path conventions is v tempting

jacobstanley:
> I guess I don't know when you'd want to reach into and resandbox a totally foreign project when you're already in one

You might have a sandbox which isn't compatible with the file you're trying to load, only the sandbox that file is supposed to be in is guaranteed to have the right packages.

jacobstanley:
@thumphries would like to discuss this next week some time if you don't mind? I'll show you my use case, perhaps we can solve it another way

thumphries:
Sure thing.

nhibberd:
:zap:

PR #158, opened 2017-02-08
mafia quick/ghci/watch -p excludes project sources

jacobstanley:
This change makes it so that project sources are not included when rebuilding dependencies as a result of invoking quick/ghci/watch -p.

This means that if you have two packages `foo-x` and `foo-y`, where `foo-y` depends on `foo-x`, and you modify `foo-x`, then invoke the `mafia quick -p` on a file in `foo-y`, it won't need to rebuild `foo-x`.

Sorry @nhibberd, not sure that I can easily fix that issue you were having where it needs to build the dependent package the first time around. The alternatives I can think of at the moment are all worse than the current behaviour.
/jury approved @charleso @thumphries @nhibberd

### DIFF
--- a/main/mafia.hs
+++ b/main/mafia.hs
@@ -420,7 +420,7 @@ mafiaDepends :: DependsUI -> Maybe PackageName -> [Flag] -> EitherT MafiaError I
 mafiaDepends ui mpkg flags = do
   lockFile <- firstT MafiaLockError $ getLockFile =<< getCurrentDirectory
   constraints <- fmap (fromMaybe []) . firstT MafiaLockError $ readLockFile lockFile
-  sdeps <- Set.toList <$> firstT MafiaInitError getSourceDependencies
+  sdeps <- fmap Set.toList . firstT MafiaInitError $ getSourceDependencies
   local <- firstT MafiaCabalError (findDependenciesForCurrentDirectory flags sdeps constraints)
   let
     deps = maybe id filterPackages mpkg $ pkgDeps local
@@ -439,7 +439,7 @@ mafiaClean = do
 
 mafiaBuild :: Profiling -> Warnings -> CoreDump -> [Flag] -> [Argument] -> EitherT MafiaError IO ()
 mafiaBuild p w dump flags args = do
-  initMafia p flags
+  initMafia LatestSources p flags
 
   let
     wargs =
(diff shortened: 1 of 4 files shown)


charleso:
💯

jacobstanley:
This is on top of https://github.com/ambiata/mafia/pull/157

nhibberd:
Best :+1:

thumphries:
👍

nhibberd:
Getting this depending on a package that has cpp in the src that I am loading in

```
["-no-user-package-db","-optP-include","-optPdist/build/autogen/cabal_macros.h","-isrc","-itest","-idist/build","-idist/build/autogen","-i../disorder-cli/src","-i../disorder-cli/test","-i../disorder-core/src","-i../disorder-core/test","-i../disorder-core/dist/build","-i../disorder-core/dist/build/autogen","-i../disorder-corpus/src","-i../disorder-corpus/test","-i../disorder-eithert/src","-i../disorder-fsm/src","-i../disorder-fsm/test","-i../disorder-jack/src","-i../disorder-jack/test","-i../disorder-lens/src","-i../disorder-lens/test","-package-db=.cabal-sandbox/x86_64-unknown-linux/7.10.2/x86_64-linux-ghc-7.10.2-packages.conf.d","src/Disorder/Aeson.hs","src/Disorder/Aeson/Time.hs"]
GHCi, version 7.10.2: http://www.haskell.org/ghc/  :? for help

../disorder-core/src/Disorder/Core/Gen.hs:28:0:
     error: missing binary operator before token "("
     #if MIN_VERSION_quickcheck_text(0, 1, 1)
     
phase `C pre-processor' failed (exitcode = 1)
```

erikd-ambiata:
@nhibberd The header file that has those package version number macros is generated by cabal at `cabal configure` time. It may be that `mafia quick` doesn't run the configure.

jacobstanley:
`mafia quick` does run configure, but it must be that because `quickcheck-text` is not a direct dependency of the package you're in, it doesn't generate the macro in the local `cabal_macros.h`?

jacobstanley:
@nhibberd I replicated this problem, and can verify that it doesn't happen in GHC 8.0 because of improvements to GHCi

The fix would be to build a similar thing for mafia, which generates the `cabal_macros.h` for all transitive dependencies instead of just the direct ones, or ignore it and wait until we're on GHC 8.0

nhibberd:
:zap:

PR #159, opened 2017-02-08
Replace other-modules with exposed-modules for Hackage packages

tranma:
Replace `other-modules` in the `library` section:

```
$ sdiff old new
library								library
  hs-source-dirs:						  hs-source-dirs:
    src								    src
  ghc-options: -Wall						  ghc-options: -Wall
  build-depends:						  build-depends:
    base >= 4.3 && < 5,						    base >= 4.3 && < 5,
    ghc-prim							    ghc-prim
  exposed-modules:						  exposed-modules:
    Data.Orphans						    Data.Orphans
  other-modules:					      |	  exposed-modules:
    Data.Orphans.Prelude					    Data.Orphans.Prelude
  default-language: Haskell2010					  default-language: Haskell2010

test-suite spec							test-suite spec
  type: exitcode-stdio-1.0					  type: exitcode-stdio-1.0
  main-is: Spec.hs						  main-is: Spec.hs
  hs-source-dirs:						  hs-source-dirs:
    test							    test
  ghc-options: -Wall						  ghc-options: -Wall
  build-depends:						  build-depends:
    base >= 4.3 && < 5,						    base >= 4.3 && < 5,
    base-orphans,						    base-orphans,
    hspec == 2.*,						    hspec == 2.*,
    QuickCheck							    QuickCheck
  other-modules:						  other-modules:
    Control.Applicative.OrphansSpec				    Control.Applicative.OrphansSpec
    Control.Exception.OrphansSpec				    Control.Exception.OrphansSpec
...
```

I'm sorry.

! @jystic 
/jury approved @amosr

### DIFF
--- a/src/Mafia/Install.hs
+++ b/src/Mafia/Install.hs
@@ -365,6 +365,9 @@ createPackageSandbox penv p@(Package (PackageRef pid _ msrc) deps _) = do
       capture (InstallUnpackFailed pid sbsrc) $
         sbcabal "unpack" ["--destdir=" <> sbsrc, renderPackageId pid]
 
+      -- Presumably we don't need to do this for local packages ever.
+      exposePackageModules srcdir
+
       Hush <- sbcabal "sandbox" ["add-source", srcdir]
 
       -- We need to shuffle anything which was unpacked to 'dist' in to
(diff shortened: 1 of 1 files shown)


amosr:
:100: 


we probably want a way to disable it just in case?

tranma:
yeah, except then some projects only builds with/without a particular mafia option, which might not be a good thing for boris.

erikd-ambiata:
As a user of mafia outside Ambiata, I don't think this is a good default. I have no problem with it as a `--savage` option, but its a really bad default.

jacobstanley:
One issue with making this optional is that it affects the packages which are kept in the global cache, which will mean that builds stop being reproducible between machines.

erikd-ambiata:
Ok, its its not optional, then I am pretty firmly against this.

charleso:
As amusing, and potentially useful as this is, this feels like a large deviation from just being just a reliable cabal. I imagine at some point we would become transparently reliant on this behaviour in some places. I don't feel as strongly as Erik, but it does feel like we should be very clear about the consequences of exposing modules we don't normally have access to, and what that means long term, before we make this change.

tranma:
I don't mind, this is more of an experiment than anything. Making it an option might help with quickly trying out things in dev but probably not much more beyond that.

jacobstanley:
This is quite challenging to make optional, I'm going to close this until we figure out a way.

I've renamed this branch to `wip/sed`.

PR #161, opened 2017-03-05
Update mafia script to remove version lock.

markhibberd:
This matches a change in usage that is more
common for mafia now. The original reasoning
for the lock, was so we could control version
breakage of mafia itself if required. At the
time we weren't sure how stable it could /
would be. In retrospect, the mafia API has
been able to remain very stable, and it has
actually (probably unsurprisingly) been the
unrelenting churn on hackage that has caused
the most issues. This led to a prefer the
latest mafia, and it promises to be backwards
compatiable approach. This means that later
versions of mafia can handle hackage breakage
and projects will keep building from checkout.

The script becomes a small version manager for
mafia. From checkout, everything just works
without a previous install of mafia:

  './mafia build'

This will install the latest version of mafia
and build your project. If you already have a
local build of mafia, it will continue to use
this, if you want to use the latest run:

  './mafia set-mafia-version'

This will build and link the latest mafia version.
If you want to run a specific git commit of mafia
you can:

  './mafia set-mafia-version f52d11e0beeedbd8ba31ad8ec3a760a8b670c138'

As it stands it removes the interactive upgrade
command on the local mafia script.

There are a couple of inline comments for somewhat 
arbitrary decisions I have made that are worth further
discussion.

### DIFF
--- a/script/mafia
+++ b/script/mafia
@@ -25,9 +25,10 @@ build_version() {
         exit "$got"
     }
     chmod +x "$MAFIA_TEMP/.cabal-sandbox/bin/mafia"
-    # Ensure it executable is on same file-system so final mv is atomic.
+    # Ensure executable is on same file-system so final mv is atomic.
     mv -f "$MAFIA_TEMP/.cabal-sandbox/bin/mafia" "$MAFIA_PATH.$$"
     mv "$MAFIA_PATH.$$" "$MAFIA_PATH" || {
+        rm -f "$MAFIA_PATH.$$"
         echo "INFO: mafia version ($MAFIA_VERSION) already exists not overiding," >&2
         echo "INFO: this is expected if parallel builds of the same version of" >&2
         echo "INFO: mafia occur, we are playing by first in, wins." >&2
--- a/script/mafia
+++ b/script/mafia
@@ -37,7 +37,7 @@ build_version() {
 }
 
 enable_version() {
-    if [ $# -eq 0 ]; then
+    if [ $# -ne 0 ]; then
         MAFIA_VERSION="$(latest_version)"
         echo "INFO: No explicit mafia version requested installing latest ($MAFIA_VERSION)." >&2
     else
@@ -57,6 +57,6 @@ exec_mafia () {
 #
 
 case "${1:-}" in
-set-mafia-version) shift; enable_version "$@" ;;
+upgrade) shift; enable_version "$@" ;;
 *) exec_mafia "$@"
 esac
--- a/script/mafia
+++ b/script/mafia
@@ -1,122 +1,61 @@
 #!/bin/sh -eu
 
 : ${MAFIA_HOME:=$HOME/.mafia}
-
-fetch_latest () {
-  if [ -z ${MAFIA_TEST_MODE+x} ]; then
-    TZ=$(date +"%T")
-    curl --silent "https://raw.githubusercontent.com/ambiata/mafia/master/script/mafia?$TZ"
-  else
-    cat ../script/mafia
-  fi
-}
+: ${MAFIA_VERSIONS:=$MAFIA_HOME/versions}
 
 latest_version () {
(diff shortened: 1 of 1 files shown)


script/mafia
```diff
@@ -1,122 +1,61 @@
...
 #
 
-if [ $# -gt 0 ]; then
-  MODE="$1"
-else
-  MODE=""
-fi
-
-case "$MODE" in
-upgrade) shift; run_upgrade "$@" ;;
```
markhibberd, line 120: Cold removing upgrade? It would be nice to reclaim that sub-command, I am not sure if this script is worth having an auto-upgraded for, you can grab a newer version pretty easily, but as it is no longer changing mafia's version it doesn't seem worth including it in the script itself.

script/mafia
```diff
@@ -1,122 +1,61 @@
...
-
-  if ! cmp $MAFIA_CUR $MAFIA_TEMP >/dev/null 2>&1; then
-    mv $MAFIA_TEMP $MAFIA_CUR
-    chmod +x $MAFIA_CUR
-    echo "New version found and upgraded. You can now commit it to your git repo."
-  else
-    echo "You have latest mafia."
-  fi
+enable_version() {
+    if [ $# -eq 0 ]; then
```
markhibberd: Not sure if I like this or not, this means `./mafia set-mafia-version` grabs the latest, it is convenient but maybe a magic 'latest' token would be better? Or a separate sub-command?

script/mafia
```diff
@@ -1,122 +1,61 @@
...
+    (cd "$MAFIA_TEMP" && ./bin/bootstrap) || {
+        got=$?
+        echo "mafia version ($MAFIA_VERSION) could not be built." >&2
+        exit "$got"
+    }
+    chmod +x "$MAFIA_TEMP/.cabal-sandbox/bin/mafia"
+    # Ensure it executable is on same file-system so final mv is atomic.
+    mv -f "$MAFIA_TEMP/.cabal-sandbox/bin/mafia" "$MAFIA_PATH.$$"
+    mv "$MAFIA_PATH.$$" "$MAFIA_PATH" || {
+        echo "INFO: mafia version ($MAFIA_VERSION) already exists not overiding," >&2
```
markhibberd: `INFO` logging token is fugly but I want it to stand out from the other output in case it ever results in surprising behaviour, happy with other suggestions.

script/mafia
```diff
@@ -1,122 +1,62 @@
...
+        got=$?
+        echo "mafia version ($MAFIA_VERSION) could not be built." >&2
+        exit "$got"
+    }
+    chmod +x "$MAFIA_TEMP/.cabal-sandbox/bin/mafia"
+    # Ensure executable is on same file-system so final mv is atomic.
+    mv -f "$MAFIA_TEMP/.cabal-sandbox/bin/mafia" "$MAFIA_PATH.$$"
+    mv "$MAFIA_PATH.$$" "$MAFIA_PATH" || {
+        rm -f "$MAFIA_PATH.$$"
+        echo "INFO: mafia version ($MAFIA_VERSION) already exists not overiding," >&2
```
markhibberd, line 32: `INFO` logging token is fugly but I want it to stand out from the other output in case it ever results in surprising behaviour, happy with other suggestions.

markhibberd:
! @jystic

jacobstanley:
This is great, not sure about the command name though, although I don't know that I have anything better.

Maybe:
```
./mafia activate latest
```
and
```
./mafia activate f52d11e
```

Open to suggestions.

erikd-ambiata:
I'm ok with `set-mafia-version`. It does what is says. With `activate` I'd need to as "activate what?".

jacobstanley:
@markhibberd I'd hate to see this get forgotten, maybe we should just keep `upgrade` as the command for now?
```
./mafia upgrade
./mafia upgrade f52d11e
```

markhibberd:
Ack, will try to push through today.

Issue #165, opened 2017-03-11
Error parsing output of `ghc-pkg list` in `hoogle` command edge case


ranweiler:
## Summary
If we have an empty package database, say, in a new cabal sandbox, invocations of `mafia hoogle` will fail when attempting to parse the string "`(no packages)`".

## Observed
Create a new project with a minimal cabal file:
```bash
mkdir example
cd example
git init  # necessary!
cabal init -n --is-library -p example
cabal sandbox init
```
And we try the `hoogle` command on a prelude function:
```
mafia hoogle 'fst'
```
We get an exit code 1 and the message "Parse failed: Invalid package: no packages".

## Expected

Exit code 0 (since our query has results) and output like:
```
Combining 74 databases
Prelude fst :: (a, b) -> a
Data.Tuple fst :: (a, b) -> a
Foreign.StablePtr deRefStablePtr :: StablePtr a -> IO a
GHC.Prim deRefStablePtr# :: StablePtr# a -> State# (RealWorld) -> (# State# (RealWorld), a #)
GHC.PrimopWrappers deRefStablePtr# :: StablePtr# a -> State# (RealWorld) -> (# State# (RealWorld), a #)
GHC.StaticPtr deRefStaticPtr :: StaticPtr a -> a
```

## Notes
I'm using GHC 8.0.1 and `cabal-install` version 1.24.0.2.

In the setting above, with an empty sandbox package database, the output of `mafia exec ghc-pkg list` will be something like:
```
/usr/local/Cellar/ghc/8.0.1_4/lib/ghc-8.0.1/package.conf.d
    Cabal-1.24.0.0
    array-0.5.1.1
    base-4.9.0.0
    # ...
    transformers-0.5.2.0
    unix-2.7.2.0
    xhtml-3000.2.1
/Users/joe/tmp/example/.cabal-sandbox/x86_64-osx-ghc-8.0.1-packages.conf.d
    (no packages)
```
This suggest we are failing [here](https://github.com/ambiata/mafia/blob/master/src/Mafia/Hoogle.hs#L55) due to not filtering enough [here](https://github.com/ambiata/mafia/blob/master/src/Mafia/Hoogle.hs#L53).

If we don't need to relate packages to databases, an alternative would be to pass the `--simple-output` flag to the [call to `hc-pkg`](https://github.com/ambiata/mafia/blob/master/src/Mafia/Hoogle.hs#L52). Even in the broken case, this would result in space-delimited output like:
```
[...]

PR #167, opened 2017-03-16
Use simple-output for hoogle pkg list

charleso:
! @jystic 

https://github.com/ambiata/mafia/issues/165

Seems reasonable enough, unless simple-output was added too recently.
/jury approved @jystic

### DIFF
--- a/src/Mafia/Hoogle.hs
+++ b/src/Mafia/Hoogle.hs
@@ -49,8 +49,8 @@ hooglePackages hackageRoot = do
   firstT MafiaInitError $ initialize Nothing Nothing
   db <- hoogleCacheDir
   hoogleExe <- installHoogle
-  Out pkgStr <- liftCabal $ sandbox "hc-pkg" ["list"]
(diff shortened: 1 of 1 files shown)


nhibberd:
Got an example of the output?

charleso:
Hmm, actually this is slightly wrong. Just want to check the build fails and will try something slightly different

charleso:
> Got an example of the output?

```
> cabal exec -- ghc-pkg list --simple-output
Cabal-1.22.4.0 HUnit-1.5.0.0 MonadRandom-0.4.2.3 QuickCheck-2.8.2 SafeSemaphore-0.10.1 StateVar-1.1.0.4 ...
```

charleso:
@jystic The build now passes. This ok by you? Is `cabal exec -- ghc-pkg list` going to the same has `sandbox hc-pkg list`?

jacobstanley:
> @jystic The build now passes. This ok by you? Is `cabal exec -- ghc-pkg list` going to the same has `sandbox hc-pkg list`?

I think so, this seems to yield the right result for the package-dbs:
```
cabal exec -- sh -c 'echo "$GHC_PACKAGE_PATH"'
```

👍

jystic:
👍

PR #168, opened 2017-03-20
Use Cabal API for determining package-id of submodules

jacobstanley:
In the early days we savagely pulled the package-id out of cabal files using a combination of `Data.Text` hacks in order to avoid pulling in `Cabal` as a dependency. Now we have `Cabal` as a dependency we might as well use the proper parser.

I was working with a `.cabal` file where our string savagery didn't work (below),  so I thought we should fix this properly.

```
package:
  foo
version:
  1.2.3
...
```

! @charleso 
/jury approved @charleso

### DIFF
--- a/src/Mafia/Cabal/Package.hs
+++ b/src/Mafia/Cabal/Package.hs
@@ -28,13 +28,16 @@ import qualified Data.Set as Set
 import qualified Data.Text as T
 
 import           Distribution.Package (Dependency(..))
+import qualified Distribution.Package as Cabal (PackageIdentifier(..))
 import qualified Distribution.Package as Cabal (PackageName(..))
 import           Distribution.PackageDescription (BuildInfo(..))
 import           Distribution.PackageDescription (CondTree(..))
 import           Distribution.PackageDescription (Executable(..))
 import           Distribution.PackageDescription (GenericPackageDescription(..))
 import           Distribution.PackageDescription (libBuildInfo)
+import qualified Distribution.PackageDescription as Cabal (GenericPackageDescription(..))
 import qualified Distribution.PackageDescription as Cabal (Library(..))
+import qualified Distribution.PackageDescription as Cabal (PackageDescription(..))
 import           Distribution.PackageDescription.Parse (ParseResult(..))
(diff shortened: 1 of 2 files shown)


src/Mafia/Cabal/Package.hs
```diff
@@ -211,33 +214,28 @@ getPackageId dir = do
...
+readPackageId :: MonadIO m => File -> EitherT CabalError m PackageId
+readPackageId file = do
+  msrc <- liftM T.unpack `liftM` readUtf8 file
+  case msrc of
+    Nothing ->
+      left $ CabalCouldNotReadPackageId file
+    Just src ->
+      case parsePackageDescription src of
+        ParseFailed err ->
+          left $ CabalFileParseError file err
```
jacobstanley, line 228: As an added bonus, we now get a specific parse error from Cabal, instead of just "could not read package-id"

src/Mafia/Cabal/Package.hs
```diff
@@ -211,33 +214,28 @@ getPackageId dir = do
...
+      liftM Just $ readPackageId f
+
+readPackageId :: MonadIO m => File -> EitherT CabalError m PackageId
+readPackageId file = do
+  msrc <- liftM T.unpack `liftM` readUtf8 file
+  case msrc of
+    Nothing ->
+      left $ CabalCouldNotReadPackageId file
+    Just src ->
+      case parsePackageDescription src of
```
charleso, line 226: Just wondering if we can/should just do this once, and pass it around. We also try to work out build tools at some point - can we share the same instance?

charleso:
👍

src/Mafia/Cabal/Package.hs
jacobstanley: Yeah it might be slightly faster, I'm not sure it's worth the complexity though, I certainly can't be bothered for millisecond gains :)

src/Mafia/Cabal/Package.hs
tmcgilchrist: Nice 👍

PR #169, opened 2017-03-22
Flip the enable_version check in script/mafia

jacobstanley:
The conditional which tests for a version number argument was flipped the wrong way around.

! @markhibberd 
/jury approved @markhibberd


markhibberd:
👍

PR #170, opened 2017-04-05
REALLY kill cabal on ctrl-c

amosr:
Now:
```
amos@Amos mafia $ mafia test
Preprocessing library ambiata-mafia-0.0.1...
Preprocessing test suite 'test' for ambiata-mafia-0.0.1...
Preprocessing test suite 'test-io' for ambiata-mafia-0.0.1...
Preprocessing test suite 'test-cli' for ambiata-mafia-0.0.1...
Running 3 test suites...
Test suite test: RUNNING...
=== prop_roundtrip_Constraint from test/Test/Mafia/Cabal/Constraint.hs:19 ===
+++ OK, passed 100 tests.
^CProcess failed: cabal test -j --show-details=streaming (exit code: -9)
```

Before:
```
amos@Amos mafia $ mafia test
Preprocessing library ambiata-mafia-0.0.1...
Preprocessing test suite 'test' for ambiata-mafia-0.0.1...
Preprocessing test suite 'test-io' for ambiata-mafia-0.0.1...
Preprocessing test suite 'test-cli' for ambiata-mafia-0.0.1...
Running 3 test suites...
Test suite test: RUNNING...
=== prop_roundtrip_Constraint from test/Test/Mafia/Cabal/Constraint.hs:19 ===
+++ OK, passed 100 tests.

^C
amos@Amos mafia $
=== prop_roundtrip_PackagePlan from test/Test/Mafia/Cabal/Dependencies.hs:20 ===
+++ OK, passed 100 tests.

=== prop_roundtrip_Flag from test/Test/Mafia/Cabal/Types.hs:19 ===
+++ OK, passed 100 tests.

^C
amos@Amos mafia $
^C
^C

=== prop_joinHooglePackages from test/Test/Mafia/Hoogle.hs:16 ===
+++ OK, passed 1 tests.

=== prop_roundtrip_PackageId from test/Test/Mafia/Package.hs:19 ===
+++ OK, passed 100 tests.

=== prop_clean from test/Test/Mafia/Process.hs:14 ===
+++ OK, passed 100 tests.

Test suite test: PASS
Test suite logged to: dist/test/ambiata-mafia-0.0.1-test.log
Test suite test-io: RUNNING...
=== prop_flock from test/Test/IO/Mafia/Flock.hs:30 ===

^C
^C
=== prop_roundtrip_PackagePlan from test/Test/Mafia/Cabal/Dependencies.hs:20 ===
+++ OK, passed 100 tests.

=== prop_roundtrip_Flag from test/Test/Mafia/Cabal/Types.hs:19 ===
+++ OK, passed 100 tests.

```
/jury approved @tranma @olorin @erikd-ambiata @jystic

### DIFF
--- a/src/Mafia/Process.hs
+++ b/src/Mafia/Process.hs
@@ -69,6 +69,7 @@ import           System.IO (IO, FilePath, Handle, BufferMode(..))
 import qualified System.IO as IO
 import qualified System.Process as Process
 import qualified System.Process.Internals as ProcessInternals
+import qualified System.Posix.Types   as Posix
 import qualified System.Posix.Process as Posix
 import qualified System.Posix.Signals as Signals
 
@@ -186,31 +187,33 @@ createProcess :: MonadIO m => Process.CreateProcess -> m (Maybe Handle, Maybe Ha
 createProcess = liftIO . Process.createProcess
 
 -- Spawn a new process, and if we get a ctrl-c, make absolutely sure everything we started is finished.
-createProcessAnnihilate :: MonadIO m => Process.CreateProcess -> m (Maybe Handle, Maybe Handle, Maybe Handle, Process.ProcessHandle)
+createProcessAnnihilate :: (MonadIO m, MonadCatch m) => Process.CreateProcess -> m (Maybe Handle, Maybe Handle, Maybe Handle, Process.ProcessHandle)
 createProcessAnnihilate cp = do
-  (a, b, c, pid) <- createProcess cp { Process.create_group = True }
+  (a, b, c, ph) <- createProcess cp { Process.create_group = True }
+  pgid <- tryProcessGroupOfProcessHandle ph
+  fromMaybe (return ()) (installInterruptHandler <$> pgid)
+  return (a, b, c, ph)
 
-  pgid <- liftIO $ ProcessInternals.withProcessHandle pid getPgid
-  case pgid of
-   Just pgid' -> do
-    _ <- liftIO $ Signals.installHandler Signals.keyboardSignal (Signals.Catch $ killer pgid') Nothing
-    return ()
-   Nothing -> return ()
+tryPosixPidOfProcessHandle :: MonadIO m => Process.ProcessHandle -> m (Maybe Posix.ProcessID)
+tryPosixPidOfProcessHandle ph =
+  liftIO $ ProcessInternals.withProcessHandle ph $
+   \case
+    ProcessInternals.OpenHandle i   -> return $ Just i
+    ProcessInternals.ClosedHandle _ -> return $ Nothing
 
-  return (a, b, c, pid)
+tryProcessGroupOfProcessHandle :: (MonadIO m, MonadCatch m) => Process.ProcessHandle -> m (Maybe Posix.ProcessGroupID)
+tryProcessGroupOfProcessHandle ph = do
+  pid <- tryPosixPidOfProcessHandle ph
+  case pid of
+   Nothing -> return Nothing
+   Just h  -> handle ignoreIOE (Just <$> liftIO (Posix.getProcessGroupIDOf h))
  where
-  getPid = \case
-   ProcessInternals.OpenHandle i -> Just i
-   ProcessInternals.ClosedHandle _ -> Nothing
+  ignoreIOE (_ :: IOException) = return Nothing
 
-  getPgid i = case getPid i of
-    Nothing -> return Nothing
-    Just h -> do
-      let ignoreIOE (_ :: IOException) = return Nothing
-      handle ignoreIOE (Just <$> Posix.getProcessGroupIDOf h)
-
-  -- SIGQUIT
-  killer = Signals.signalProcessGroup Signals.keyboardTermination
+installInterruptHandler :: MonadIO m => Posix.ProcessGroupID -> m ()
+installInterruptHandler pgid = do
+  _ <- liftIO $ Signals.installHandler Signals.keyboardSignal (Signals.Catch $ Signals.signalProcessGroup Signals.keyboardTermination pgid) Nothing
+  return ()
 
 
 class ProcessResult a where
--- a/src/Mafia/Process.hs
+++ b/src/Mafia/Process.hs
@@ -206,7 +206,9 @@ tryProcessGroupOfProcessHandle ph = do
   pid <- tryPosixPidOfProcessHandle ph
   case pid of
    Nothing -> return Nothing
-   Just h  -> handle ignoreIOE (Just <$> liftIO (Posix.getProcessGroupIDOf h))
+   Just h  -> handle ignoreIOE $ do
+    pgid <- liftIO (Posix.getProcessGroupIDOf h)
+    return $ Just pgid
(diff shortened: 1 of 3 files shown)


amosr:
! @jystic

tranma:
👍 👍 👍 because I had to `killall -9` every time anyway, this is doing the same

src/Mafia/Process.hs
```diff
@@ -174,6 +182,36 @@ renderProcessError = \case
...
+
+-- Spawn a new process, and if we get a ctrl-c, make absolutely sure everything we started is finished.
+createProcessAnnihilate :: MonadIO m => Process.CreateProcess -> m (Maybe Handle, Maybe Handle, Maybe Handle, Process.ProcessHandle)
+createProcessAnnihilate cp = do
+  (a, b, c, pid) <- createProcess cp { Process.create_group = True }
+
+  pgid <- liftIO $ ProcessInternals.withProcessHandle pid getPgid
+  case pgid of
+   Just pgid' -> do
+    _ <- liftIO $ Signals.installHandler Signals.keyboardSignal (Signals.Catch $ killer pgid') Nothing
```
olorin: Sidenote, `keyboardSignal` has to be the worst alias for `SIGINT` I've ever seen.

olorin:
I'm not sure how I feel about sending `KILL` as part of normal operation - if `cabal`'s behaviour isn't totally braindead we shouldn't need to. How does it respond to `TERM` and `QUIT`?

erikd-ambiata:
Yeah, I'd be inclined to send `SigQuit` first, waiting say 10 seconds and then sending `SigKill`.

olorin:
> Yeah, I'd be inclined to send SigQuit first, waiting say 10 seconds and then sending SigKill.

That'd be preferable; I'd probably still only do it after establishing that `cabal` frequently didn't respond to `QUIT`, I'm not sure if that's been tried?

tranma:
But I (the user) don't want to wait 10 seconds :)

amosr:
> if cabal's behaviour isn't totally braindead we shouldn't need to.

Well…

Cool, SIGQUIT does work!

olorin:
:100:

markhibberd:
Mmmm, does this break ghci ctrl-c? that would be bad

amosr:
> Mmmm, does this break ghci ctrl-c? that would be bad

aren't `mafia repl` and `cabal repl` already broken? I'll enable it just for `mafia test` and `mafia bench`.

erikd-ambiata:
:+1: Proper Ctrl-C handling on test is the important one.

markhibberd:
> aren't mafia repl and cabal repl already broken?

It very well could be, I guess I rarely use it over calling quick or ghci directly. And given I hit ctrl-c a lot and haven't noticed any issues might say something. But the ability to quit something running in ghci is highly valuable, would be good to not dissallow it, if cabal repl could be fixed.

jacobstanley:
@amosr I wonder if we could fix `mafia repl` in a similar way to what you've done here, install a signal handler, then pass the `SIGINT` to the child `ghci` rather than giving it to `cabal`

src/Mafia/Process.hs
```diff
@@ -174,6 +182,37 @@ renderProcessError = \case
 
 ------------------------------------------------------------------------
 
+createProcess :: MonadIO m => Process.CreateProcess -> m (Maybe Handle, Maybe Handle, Maybe Handle, Process.ProcessHandle)
+createProcess = liftIO . Process.createProcess
+
+-- Spawn a new process, and if we get a ctrl-c, make absolutely sure everything we started is finished.
+createProcessAnnihilate :: MonadIO m => Process.CreateProcess -> m (Maybe Handle, Maybe Handle, Maybe Handle, Process.ProcessHandle)
+createProcessAnnihilate cp = do
+  (a, b, c, pid) <- createProcess cp { Process.create_group = True }
```
jacobstanley: what does `create_group = True` do?

markhibberd: I am pretty sure it just calls `setsid`

src/Mafia/Process.hs
```diff
@@ -174,6 +182,37 @@ renderProcessError = \case
...
+   ProcessInternals.ClosedHandle _ -> Nothing
+
+  getPgid i = case getPid i of
+    Nothing -> return Nothing
+    Just h -> do
+      let ignoreIOE (_ :: IOException) = return Nothing
+      handle ignoreIOE (Just <$> Posix.getProcessGroupIDOf h)
+
+  -- SIGQUIT
+  killer = Signals.signalProcessGroup Signals.keyboardTermination
```
jacobstanley: Any reason `getPid`, `getPgid` and `killer` shouldn't just be top level functions? And in that case maybe `killer` should be called `sigquit` or something to that effect?

amosr:
> I wonder if we could fix mafia repl in a similar way to what you've done here, install a signal handler, then pass the SIGINT to the child ghci rather than giving it to cabal

that'd be good, but I couldn't find a way to get the list of children for a pid (without parsing the output of `ps`): something like

```
children :: ProcessId -> IO [ProcessId]
```

so that's why it creates a process group - sending quit to the process group should get cabal's children as well.

I might be able to move cabal to a new process group before sending ctrl-c, which would mean it'd only send it to the children

amosr:
nah, I couldn't figure out how to fix the repl. perhaps someone wants to fix this in cabal itself, but not me

erikd-ambiata:
If someone can give me a reproducible test case, i might have a look at fixing it in cabal.

amosr:
I guess there are two things, which should be reproducible on any .cabal project with tests:

1. cabal test should not leave things running after ^C:
```
$ ps | grep setup
48923 ttys008    0:00.00 grep --color=always setup

$ cabal test
Test suite test-io: RUNNING...
^C

$ ps | grep setup
48961 ttys008    0:00.13 /blahblahblah/dist/setup/setup test --verbose=1 --builddir=dist --log=$pkgid-$test-suite.log --machine-log=$pkgid.log --show-details=failures test test-io test-cli
49303 ttys008    0:00.00 grep --color=always setup
```

2. cabal repl should not die after ^C: if you send ^C to normal ghci it clears the line: 
```
$ ghci
GHCi, version 7.10.2: http://www.haskell.org/ghc/  :? for help
Prelude
λ ^C
Prelude
λ ^C
Prelude
λ
```

but if you send ^C to cabal repl, it dies:

```
$ cabal repl
*Mafia.Bin
λ ^C
*Mafia.Bin
λ
<stdin>: hGetChar: hardware fault (Input/output error)

$ 
```

jacobstanley:
@erikd the cabal repl one should be easy to fix, cabal just needs to exec ghci rather than whatever it's doing now, I suspect it's not using exec because windows

jacobstanley:
👍 good with this as is on 🍏 

The `cabal repl` thing is a separate issue

jacobstanley:
> but if you send ^C to cabal repl, it dies:

The best part about this is that it leaves GHCi running in the background!

amosr:
forgot about this. merging now

PR #171, opened 2017-04-21
Add finite backjump limit

olorin:
No limit by default is problematic because it makes identifying
nonterminating searches require more effort ("dependency resolution is
taking ages, is it just really slow due to hardware/load/gremlins or
is there another problem?")

I think this should be configurable; I made a start on implementing
this in `topic/term2`, but it's not ready yet so I thought I would put
this up in the interim.

! @nhibberd @jystic

### DIFF
--- a/src/Mafia/Cabal/Dependencies.hs
+++ b/src/Mafia/Cabal/Dependencies.hs
@@ -190,7 +190,10 @@ makeInstallPlan mdir sourcePkgs installArgs = do
       installDryRun args =
         cabal "install" $
           [ "--reorder-goals"
-          , "--max-backjumps=-1"
+            -- 10^3 is an arbitrary "should be enough for anyone" number,
+            -- used rather than -1 to avoid running forever in build scripts
+            -- et cetera.
+          , "--max-backjumps=1000"
           , "--avoid-reinstalls"
           , "--dry-run" ] <> installArgs <> args
 


olorin:
Anyone care to comment?

! @ambiata/engineers

nhibberd:
I think waiting until next week would be the best option

olorin:
Happy to wait until next week, figured if I tag a random set of people today they'd see it Monday morning (or whenever, this isn't urgent though it's a bit annoying to work around).

erikd-ambiata:
I can't see any downsides to this, but leave the final decision to @jystic .

jacobstanley:
Have you actually observed this breaking in practice?

`1000` is a really low default, it's actually even lower than what `cabal` uses by [default](https://github.com/haskell/cabal/blob/d119a02cbbd44c91892d4e6a2e11c0c1836eea87/cabal-install/Distribution/Client/Setup.hs#L1498) if you don't set anything.

`-1` is what is used by the stack [solver mode](https://github.com/commercialhaskell/stack/blob/5f7c0f2914c38beb212bec0704fcfe94940333db/src/Stack/Solver.hs#L122) and was the only thing I could get to work consistently with all of our projects.

If we have to set it I'd prefer something much larger, like `100000`, but what do we do in the situation that a project needs even more than that to solve properly?

olorin:
> Have you actually observed this breaking in practice?

Yep, wouldn't have touched it otherwise. Would you be okay with leaving the default as is but adding an option for the user to override?

jacobstanley:
> Would you be okay with leaving the default as is but adding an option for the user to override?

Yeah that sounds good.

I'm ok having a limit which isn't `-1` but I think we need to do some testing to make sure it works with everything. Perhaps it's safer / easier to just leave it and have the user option though, as you suggest.

tmcgilchrist:
@jystic I've had a few cases of non-terminating searches in personal projects, having an option to change it would have helped. I ended up hand crafting a lock file to get termination and then slowly modifying that until I had better package version constraints.

olorin:
Will reopen with configurable version.

PR #174, opened 2017-05-10
Topic/split readme

erikd-ambiata:
First steps in making documentation better for external users. Changes to existing docs are minor, other than splitting them into separate linked documents.

! @jystic @tmcgilchrist 
/jury approved @jystic


tmcgilchrist:
What's the reason for splitting these up? Assuming there is another step to make the docs prettier or available as a github page (other doc serving site)/

erikd-ambiata:
A couple of reasons. Firstly the main doc is getting a bit unweildy, mixing, building, using and command reference material. Secondly in its present form its difficult to add something like a Getting Started guide.

jacobstanley:
I'm ok with the splitting, but I'm not sure what is new here and what is just splitting?

erikd-ambiata:
The only thing that is new is the boostrapping document. Everything else is splitting with one minor tweak (adding "a light weight but" to "opinionated wrapper around Cabal").

jacobstanley:
I would +1 without the bootstrapping doc, I think that should be a separate PR so people can comment

erikd-ambiata:
WIll drop the bootstrapping doc.

jacobstanley:
👍

PR #185, opened 2017-10-09
Update ghcid to handle 8+ CallStack changes.

markhibberd:
Without this, running commands with mafia watch results
in call stacks being printed from how ghcid escapes out
of ghci with 'error' calls.

Having these versions hardcoded is less than ideal, it 
assumes people have the same hackage index and the
same version will always work across versions which is
not true. Leaving that as is for the moment, but I will 
update if this new version causes issues anywhere.


markhibberd:
For reference: https://github.com/ndmitchell/ghcid/commit/714f3a50c54fcaeb1a4f6408f260ed65411bcba7

charleso:
Yeah the hard-code version was my dumb idea. :(

charleso:
Thanks Mark. 👍

PR #188, opened 2017-11-13
Bump opt parse upper bound.

tmcgilchrist:
! @jystic 
/jury approved @jystic


tmcgilchrist:
Matches the ranges in [x-optparse](https://github.com/ambiata/x/blob/master/x-optparse/ambiata-x-optparse.cabal)

jacobstanley:
👍

PR #190, opened 2017-12-04
First PR for cabal-2.0 support

erikd-ambiata:
Hacking this to make it work made a rather large mess. Going to split this up as a set of smaller PRs.

! @jystic @tmcgilchrist 
/jury approved @jystic

### DIFF
--- a/src/Mafia/Cabal/Dependencies.hs
+++ b/src/Mafia/Cabal/Dependencies.hs
@@ -225,26 +225,38 @@ takeReinstall p =
 parseInstallPlan :: Text -> Either CabalError [PackagePlan]
 parseInstallPlan =
   first (CabalInstallPlanParseError . T.pack) .
-  traverse parsePackagePlan .
-  List.drop 1 .
-  List.dropWhile (/= "In order, the following would be installed:") .
-  T.lines
+  A.parseOnly pPackagePlans
 
 parsePackagePlan :: Text -> Either String PackagePlan
 parsePackagePlan txt =
-  let go err = "Invalid package plan: " <> T.unpack txt <> "\nExpected: " <> err
+  let go err = "Invalid package plan: '" <> T.unpack txt <> "'\nExpected: " <> err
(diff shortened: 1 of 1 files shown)


erikd-ambiata:
! @jystic @tmcgilchrist This PR only changes the package plan parser so that it is *capable* of parsing the cabal-2.0 package plan output. Other than changing the this parser it has no other functional changes.

src/Mafia/Cabal/Dependencies.hs
```diff
@@ -293,9 +304,18 @@ pPackageChange =
...
+pPackageId :: Parser PackageId
+pPackageId = do
+  let
+    isPkgIdChar c
+        | c >= 'a' && c <= 'z' = True
+        | c >= 'A' && c <= 'Z' = True
+        | c >= '0' && c <= '9' = True
+        | c == '-' = True
+        | c == '.' = True
+        | otherwise = False
```
jacobstanley, line 316: Curious where you found these rules? (just curious)

jacobstanley:
👍 looks ok to me, I'm assuming you've tested this out with the cabal 1.24 and everything still works

src/Mafia/Cabal/Dependencies.hs
erikd-ambiata: Its from memory. I read it in the cabal sources at some point. At the moment google can only find : https://github.com/commercialhaskell/stack/blob/master/src/Stack/Types/PackageName.hs#L76-L84

erikd-ambiata:
> I'm assuming you've tested this out with the cabal 1.24 and everything still works

Yep, I was running mafia with this and some other patches and then switching back and forth between cabal-install-1.24 and 2.0 while I was fiddling with https://github.com/ambiata/cabal-extras/pull/68 and https://github.com/ambiata/parsley/pull/37 .

tmcgilchrist:
Nice work!

PR #191, opened 2017-12-04
Support cabal >= 1.22.4 && < 2.1 

erikd-ambiata:
On top of #190

! @jystic @tmcgilchrist 
/jury approved @tmcgilchrist

### DIFF
--- a/ambiata-mafia.cabal
+++ b/ambiata-mafia.cabal
@@ -96,6 +96,14 @@ executable mafia
   main-is:
                     ../main/mafia.hs
 
+  other-modules:
+                    BuildInfo_ambiata_mafia
+                    DependencyInfo_ambiata_mafia
+
+  autogen-modules:
+                    BuildInfo_ambiata_mafia
+                    DependencyInfo_ambiata_mafia
+
   build-depends:
                       base
                     , ambiata-mafia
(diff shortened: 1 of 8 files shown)


tmcgilchrist:
☕️

erikd-ambiata:
Anyone who upgrades to this version of mafia will be fine as long as they keep their cabal-install version at 1.24.*.

If they upgrade to cabal-install version 2.0, and build new stuff they will also be fine. However, if the try to build much of our existing code it will fail because the `Setup.hs` files in most of our projects need to be upgraded to the one that is currently in the cabal-extras package.

Making out stuff work with cabal-install version 2.0 requires that we update the `Setup.hs` files in all the projects. The version of `Setup.hs` in cabal-extras already builds with the Cabal 2.0 library, but we should also make sure it builds with ghc 8.2 while we are at it.

erikd:
This now builds in Travis with three versions of GHC and two versions of cabal-install (6 combinations).

ambiata-mafia.cabal
```diff
@@ -22,7 +22,7 @@ library
                     , attoparsec                      >= 0.12       && < 0.14
                     , base16-bytestring               == 0.1.*
                     , bytestring                      == 0.10.*
-                    , Cabal                           >= 1.22       && < 1.25
+                    , Cabal                           >= 2.0        && < 2.1
```
tmcgilchrist, line 25: This lower bound seems like it would stop use of Cabal 1.24.0.*?

erikd: Yes, code currently doesn't compile against the old version of the library. Had to do this to avoid problems with `Setup.hs` and the fact that GHC 7.10 doesn't support version macros in that file.

tmcgilchrist (approved):

PR #192, opened 2017-12-05
master.toml: Cache mafia builds

erikd-ambiata:
! @jystic @tmcgilchrist

### DIFF
--- a/master.toml
+++ b/master.toml
@@ -5,40 +5,46 @@
 
 [build.dist-7-8]
   GHC_VERSION="7.8.4"
-  CABAL_VERSION="1.22.4.0"
+  CABAL_VERSION="1.24.0.0"
   MAFIA_HAPPY = "true"
+  CACHE = "true"
 
 [build.dist-7-10]
   GHC_VERSION="7.10.2"
-  CABAL_VERSION="1.22.4.0"
+  CABAL_VERSION="1.24.0.0"
   MAFIA_HAPPY = "true"
   HADDOCK = "true"
   HADDOCK_S3 = "$AMBIATA_HADDOCK_MASTER"
   PUBLISH = "true"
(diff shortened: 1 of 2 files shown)


master.toml
```diff
@@ -5,40 +5,46 @@
...
 [build.dist-8-0]
   GHC_VERSION="8.0.1"
   CABAL_VERSION="1.24.0.0"
   MAFIA_HAPPY = "true"
+  CACHE = "true"
 
 [build.branches-7-8]
   GHC_VERSION = "7.8.4"
-  CABAL_VERSION = "1.22.4.0"
+  CABAL_VERSION = "1.24.0.0"
```
tmcgilchrist, line 31: Does that version of cabal work with GHC 7.8.4?

tmcgilchrist:
If it builds sure 👍

master.toml
erikd-ambiata: I though it does. We'll see soon enough. If not, I'll remove caching for the 7.8 build.

erikd-ambiata:
Mafia builds in CI using a `cabal` script that is stored in the repo. Not sure if caching makes sense. 

@jystic ?

master.toml
jacobstanley: caching doesn't work for 7.8 anyway

jacobstanley:
> Mafia builds in CI using a cabal script that is stored in the repo. Not sure if caching makes sense.

I think actually mafia is built using mafia because of the master runner which is in use.

jacobstanley:
Ah, having said that, the master runner it's using doesn't support caching as it is too old

jacobstanley:
I think you might potentially run in to a chicken and egg problem if you upgrade the master runner, not sure

erikd-ambiata:
Updated the master runner to see if it works.

PR #194, opened 2017-12-07
[Do not merge]: Drop special case handling of lib directory

erikd-ambiata:
I can't see any reason why the `lib` directory should be special. I'm also using mafia to build a project that has a cabal file in the `lib/` directory.

! @jystic @tmcgilchrist

### DIFF
--- a/src/Mafia/Submodule.hs
+++ b/src/Mafia/Submodule.hs
@@ -136,7 +136,6 @@ getSourcesFrom dir = do
            . fmap   (takeDirectory)
            . filter (not . T.isInfixOf ".cabal-sandbox/")
            . filter (not . T.isPrefixOf "dist-newstyle/")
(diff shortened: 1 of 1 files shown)


tmcgilchrist:
Can't offer anything here I don't have context on why this was here in the first place.

erikd-ambiata:
Investigating why tests are failing.

erikd-ambiata:
Quizzed @jystic about the `lib/` directory thing.

The function `getSourcesFrom` lists all the files in the submodules of the project mafia is building. It excludes the `lib/` because we want to exclude the submodules of the submodules.

Since we still need to list the files in all the submodules, we need to handle this a different way.

charleso:
One generic option is to run `git submodule` in each submodule and exclude anything from that list. Not sure if there's an easier way.

PR #195, opened 2017-12-20
Inline Twine.

tmcgilchrist:
Thought exercise in making twine an inline dependency rather than a git submodule.
Not the most minimal version of twine possible, just a working build of it.

The other option would be to make twine a real package on hackage and depend on that.

I have made a start on removing the submodules in twine here https://github.com/tmcgilchrist/twine/tree/topic/either 

The remaining issue is the use of `P` and whether that is going up on hackage or just inline a minimal version.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "lib/disorder"]
-	path = lib/disorder
-	url = https://github.com/ambiata/disorder.hs
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,4 +4,6 @@
 [submodule "lib/x"]
 	path = lib/x
 	url = https://github.com/haskell-mafia/x
-
+[submodule "lib/p"]
+	path = lib/p
+	url = git@github.com:ambiata/p.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,6 +7,4 @@
 [submodule "lib/x"]
 	path = lib/x
 	url = https://github.com/haskell-mafia/x
-[submodule "lib/twine"]
-	path = lib/twine
-	url = https://github.com/haskell-mafia/twine
+
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "lib/p"]
-	path = lib/p
-	url = https://github.com/ambiata/p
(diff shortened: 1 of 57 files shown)


erikd-ambiata:
I'm cool with this. I think its important to get mafia to a stage where people can just cabal install it.

As for `p` I think mafia should just use the standard prelude and the stuff in `x` does mafia use anything that is not in `transformers-either`? Finally, `disorder` should just be swapped out for hedgehog.

👍

src/Mafia/Twine/Async.hs
```diff
@@ -0,0 +1,78 @@
...
+
+import           Mafia.Twine.Snooze
+
+import           System.IO (IO)
+
+import           Control.Monad.Trans.Either
+
+data AsyncTimeout =
+  AsyncTimeout Duration
+  deriving (Eq, Show)
```
erikd, line 30: Any reason for that not to be a `newtype`?

.gitmodules
```diff
@@ -1,12 +1,9 @@
...
 	url = https://github.com/ambiata/disorder.hs
 [submodule "lib/x"]
 	path = lib/x
 	url = https://github.com/ambiata/x
-[submodule "lib/twine"]
-	path = lib/twine
-	url = https://github.com/ambiata/twine
+[submodule "lib/p"]
+	path = lib/p
+	url = git@github.com:ambiata/p.git
```
erikd: P still needed as a submodule? What about X?

src/Mafia/Twine/Async.hs
tmcgilchrist: I'm purposefully not improving / changing twine beyond what is absolutely necessary to get mafia working. There's a good case for either making twine a real thing on hackage or improving the twine inlined here (I'd be voting for the first option).

.gitmodules
tmcgilchrist: There is still work in progress to remove all the submodules, first with x-optparse and then anything in disorder. Only then can i remove p entirely.

erikd-ambiata: So probably should remove all these submodules in one go.

erikd (approved):

PR #196, opened 2017-12-23
Add travis build

erikd:
Only builds with GHC 7.10 and 8.0 for now. Can't build in Travis with 7.8.4 because the compiler available hasn't been built with profiling and hence the tests fail. Building with 8.2 is also busted and will be fixed in an upcoming PR.

@haskell-mafia

### DIFF
--- a/src/Mafia/Script.hs
+++ b/src/Mafia/Script.hs
@@ -5,11 +5,20 @@ module Mafia.Script (
     ScriptError(..)
   , renderScriptError
   , runScript
+
+  -- For testing.
+  , Pragma (..)
+  , Submodule (..)
+  , SubmoduleLocation (..)
+  , SubmoduleProtocol (..)
+  , pSubmodule
+  , renderPragmaSubmodule
   ) where
 
 import           Control.Monad.IO.Class (MonadIO(..))
 
 import qualified Data.Attoparsec.Text as Atto
+import           Data.Char (isSpace)
 import qualified Data.Map as Map
 import qualified Data.Text as T
 
@@ -41,11 +50,23 @@ newtype Package =
       unPackage :: Text
     } deriving (Eq, Ord, Show)
 
+data SubmoduleProtocol =
+    Https
+  | Git
+    deriving (Eq, Ord, Show)
+
 data Submodule =
   Submodule {
-      submoduleUser :: !Text
+      submoduleLocation :: !(Maybe SubmoduleLocation)
+    , submoduleUser :: !Text
     , submoduleRepo :: !Text
     , submoduleCommit :: !(Maybe Text)
+    } deriving (Eq, Ord, Show)
+
+data SubmoduleLocation =
+  SubmoduleLocation {
+      submoduleProtocol :: !SubmoduleProtocol
+    , submoduleService :: !Text
     } deriving (Eq, Ord, Show)
 
 data Pragma =
@@ -82,9 +103,9 @@ pPackage =
   fmap PragmaPackage . pPragma "PACKAGE" $
     Package . T.strip <$> Atto.takeWhile (/= '#')
 
-pAlphaNum :: Atto.Parser Text
-pAlphaNum =
-  Atto.takeWhile1 (Atto.inClass "a-zA-Z0-9")
+pNameChar :: Atto.Parser Text
+pNameChar =
+  Atto.takeWhile1 (Atto.inClass "a-zA-Z0-9-")
 
 pHash :: Atto.Parser Text
 pHash =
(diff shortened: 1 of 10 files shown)


erikd:
Ok, I finally have something that builds with ghc 7.8.4, 7.10.3 and 8.0.2.

Unfortunately the builds sometimes fail when building under travis. This can probably fixed by simplifying some of the tests.

src/Mafia/Script.hs
```diff
@@ -136,7 +136,7 @@ parseScript path source =
 
 renderSubmoduleUrl :: Submodule -> Text
 renderSubmoduleUrl s =
-  "git@github.com:" <> submoduleUser s <> "/" <> submoduleRepo s
+  "https://github.com/" <> submoduleUser s <> "/" <> submoduleRepo s
```
nhibberd: Isn't this a breaking change for mafia scripts? can the protocol be added to the Pragma in a compatible way? 

e.g.

```
data Submodule =
  Submodule {
      submoduleProtocol :: !Protocol
    , submoduleUser :: !Text
    , submoduleRepo :: !Text
    , submoduleCommit :: !(Maybe Text)
    } deriving (Eq, Ord, Show)

Current implicit behaviour:
{-# SUBMODULE ambiata/p@xxx #-} -> Submodule Ssh "ambiata" "p" "xxx"

Explicit protocol:
{-# SUBMODULE ssh ambiata/p@xxx #-} -> Submodule Ssh "ambiata" "p" "xxx"

{-# SUBMODULE https ambiata/p@xxx #-} -> Submodule Https "ambiata" "p" "xxx"
```

I am sure there is a better way of describing the protocols, just first thing that came to mind that would keep compatibility

erikd: Yep, that makes sense. Will look into it.

erikd: Actually, why not make the protocol and the server explicit like:
```
{-# SUBMODULE git@github.com ambiata/p@xxx #-}
{-# SUBMODULE https://gitlab.com ambiata/x@xxx #-}
```
so that other services like Gitlab and Bitbucket can be used.

Its harder to make this backwards compatible though.

HuwCampbell:
Just throwing in something slightly orthogonal.

I've wanted to explicitly specify an sdist tarball which isn't on hackage as a submodule. So extensions might be interesting to consider too.

erikd:
@HuwCampbell By all mean raise an issue, but this PR is about getting mafia building in travis without breaking any current functionality.

erikd:
New version allows:

```
{-# SUBMODULE ambiata/p@xxx #-}      
{-# SUBMODULE git@github.com ambiata/p@xxx #-}  
{-# SUBMODULE https://gitlab.com ambiata/x@xxx #-}
```
where the first assumes the location is `git@github.com` and is therefore backwards compatible with what we had.

Also added a property test to round-trip those `SUBMODULE` lines.

src/Mafia/Script.hs
```diff
@@ -5,11 +5,20 @@ module Mafia.Script (
     ScriptError(..)
   , renderScriptError
   , runScript
+
+  -- For testing.
+  , Pragma (..)
```
charleso, line 10: This whole script change seems unrelated to building in travis. Should it be it's own PR?

src/Mafia/Script.hs
charleso: > Actually, why not make the protocol and the server explicit like:

I'm actually 100% on board with this, I never liked hard coding to github. That said, my suggestion is that you should also fall back to the previous behaviour to remain backwards compatible (as well as keeping with Jacob's original intention)

charleso:
I am curious why not keep the original ambiata submodules for disorder/x/twine rather than forking them? Especially if the plan is to just inline them at some point. Are the forks different?

charleso:
I'm 👍 for a travis build, but the script change seems like it should be kept separate _and_ remain backwards compatible.

erikd:
I forked the ambiata submodules because as they were, they would not clone over a HTTPS transport, but required SSH transport.

The idea is to get this building in travis and then remove the submodules so that this can then be put in Hackage.

charleso:
> I forked the ambiata submodules because as they were, they would not clone over a HTTPS transport, but required SSH transport.

"required ssh"? I'm curious why not - what stops it being cloned (that's news to me)?

charleso:
> what stops it being cloned (that's news to me)?

Ahh sorry you mean their submodules are ssh and not http? I'm with you now. Sorry about that.

erikd:
Yeah, the script tests needed changes because they *required* SSH access to clone the required repos.

I'm pretty sure this is the absolute minimum set of changes to get travis builds working :).

Edit: Oh, and the Script change is backwards compatible :)

tmcgilchrist (approved):
Getting this building on travis is a good first step. 
We'll inline / remove the submodules in later PRs

erikd:
@nhibberd @charleso Ok with this? You'll have to acknowldge via an actual "review".

PR #197, opened 2017-12-24
Remove boris builds

nhibberd:
https://github.com/haskell-mafia/mafia/pull/196


erikd:
Yeah, once #196 works this should build and pass tests.

erikd (approved):
Good to go once this has been rebased on top of #196 .

PR #203, opened 2018-01-19
Make most git refs to `ambiata` repo point to `haskell-mafia`

justanotherdot:
Ambiata has forked `mafia` from `haskell-mafia` following build failures
on GHC 7.10; this makes sure `haskell-mafia` won't get weird build
failures by ensuring that any git references should be to the new
`haskell-mafia` project instead of to `ambiata`.


justanotherdot:
Pinging @erikd

tmcgilchrist (approved):

justanotherdot:
Thanks @tmcgilchrist

PR #204, opened 2018-01-20
Fix bin/bootstrap (./lib no longer exists)

mitchellwrosen:


tmcgilchrist:
Looks good to me. Thanks

mitchellwrosen:
Of course! And if you don't mind clarifying, what is the purpose of `mafia` now that `cabal` itself has a nix-style store? A co-worker suggested I check out `mafia` as an alternative to `stack` so, well, here I am :)

erikd:
Mafia's purpose was to provide a good build tool to be used inside Ambaita. As you can see from the git log, it was started in June 2015 and was probably usable a month or two later. All of this is long before cabal new-build was ready.

Mafia was always a *thin* wrapper around cabal. I suspect that as cabal new-build improves Mafia is likely to become thinner and thinner and maybe even become completely redundant.

That co-worker wouldn't happen to be @thumphries would it?

erikd (approved):

Issue #208, opened 2018-02-23
mafia tries to get non-cabal build tools


domdere:
I'm getting bulids failing with this during the "build dependencies" phase:

```
Detected 'unzip' was required to build zip-archive-0.3.2.4
cabal: There is no package named 'unzip'.
You may need to run 'cabal update' to get the latest list of available
```

`zip-archive` has the following lines in its `.cabal` file:

```
Build-tools:     unzip
```

I already had `unzip` on my machine and it was trying to install it from hackage (and its not there obviously)

erikd:
I think that mafia assumes that all build tools listed in a cabal file are tools that are "cabal installable", like `hsc2hs` and `markdown-unlit`. Obviously this is not true for `unzip`. 

I think mafia always installing build tools listed in the cabal file is a mistake. It should first try to see the required tool exists on the `$PATH`. This would mean it would look for tools like `hsc2hs` on the `$PATH` first and only install it if its not there. This would fix your use case for `unzip` assuming that `unzip` is already on the `$PATH` but still give confusing error messages for non-haskell tools that are not installed.

Wanna send a patch?

domdere:
yeah would love to, be a little while till i find the time to do so though

domdere:
hmm bit of a problem with the suggested solution though, if it already exists in the `PATH` but *is* cabal installable, and the version is older and doesnt satisfy the constraints, then a valid version will not get installed.

I think i'll change it so it does attempt to install it from hackage, if that fails and it already exists on the `PATH` I wont make a it an error for the build

erikd:
Damn! Just ran into this problem myself. I may beat you to it!

tmcgilchrist:
Duplicate of https://github.com/haskell-mafia/mafia/pull/209

amosr:
Haha! I just ran into this morning and didn't see this issue. Interesting that this feature has been in Mafia for a while, but we all ran into it within two weeks of each other.

erikd:
I think listing non-haskell build tools in that stanza is a new thing.

PR #209, opened 2018-03-08
Continue building after failing to install build-tool

amosr:
Some packages refer to build-tools that are not cabal packages.
For example, "zip-archive" depends on "unzip", but this is talking about a binary, not a cabal package.
I think cabal allows anything in the build-tools field. Since build-tools has existed in the wild without any tool checking, there are probably more packages like this.

This is unfortunate, and perhaps the package shouldn't include this as it isn't a build tool, but we should probably provide some way to continue building regardless. Here I just catch the error and try to continue.

### DIFF
--- a/src/Mafia/Install.hs
+++ b/src/Mafia/Install.hs
@@ -379,12 +379,28 @@ installBuildTools :: CacheEnv -> Set BuildTool -> EitherT InstallError IO [Direc
 installBuildTools env tools = do
   pkgs <-
     for (Set.toList tools) $ \(BuildTool name constraints) ->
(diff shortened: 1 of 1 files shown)


amosr:
(Unfortunately I still can't build `zip-archive` locally, but I think that's unrelated…)

erikd:
Squash that into a single commit and I think its ready to go.

amosr:
@erikd squashed.

I don't think this is the *best* way to solve this – I think your suggestion in #208 to check on the PATH first is better, but I think this is better than nothing

erikd (approved):

PR #214, opened 2018-03-28
Update cabal install doc to include 2.0.0.1

justanotherdot:
Possibly not quite complete. Should we:

* Simply specify `mkdir $HOME/bin` instead of the comment re: assumption?
* Update versioning info for the library doc segment?
* ~~The GHC install guides plunk `ghc` into `$HOME/haskell`, but the lock down comment mentions `$HOME/.ghc`, does this need to be updated as well?~~

@tmcgilchrist

### DIFF
--- a/docs/installing-cabal.md
+++ b/docs/installing-cabal.md
@@ -37,10 +37,6 @@ didn't use the `bootstrap.sh` script, then make sure you run `cabal
 --version` to double check the version of the Cabal library that you
 need.
 
-*GHC 8.0.1 ships with Cabal-1.24.0.2 out of the box, so you may already
-have it installed. In any case, there is no harm in performing this step
-redundantly.*
-
 ```sh
 # double-check that we're installing the right version of the Cabal library
 $ cabal --version
--- a/docs/installing-cabal.md
+++ b/docs/installing-cabal.md
@@ -5,22 +5,23 @@ Cabal comes in two parts, the `cabal` executable (which is in the
 cabal-install package) and the Cabal library (which is in the Cabal
 package).
 
-#### Install cabal-install 1.24.0.0
+#### Install cabal-install 1.24.0.2/2.0.0.1
 
 These instructions copy the `cabal` executable to `$HOME/bin` which is
 assumed to be on your `$PATH`. If you have an alternative location where
 you prefer to put executables, then use that instead.
 
 ```sh
-$ wget http://hackage.haskell.org/package/cabal-install-1.24.0.0/cabal-install-1.24.0.0.tar.gz
-$ tar xf cabal-install-1.24.0.0.tar.gz
-$ cd cabal-install-1.24.0.0
+$ VER=1.24.0.2 # or 2.0.0.1 if using 8.2.2 or higher.
+$ wget https://hackage.haskell.org/package/cabal-install-${VER}/cabal-install-${VER}.tar.gz
+$ tar xf cabal-install-${VER}.tar.gz
+$ cd cabal-install-${VER}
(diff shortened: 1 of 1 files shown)


docs/installing-cabal.md
```diff
@@ -5,16 +5,17 @@ Cabal comes in two parts, the `cabal` executable (which is in the
 cabal-install package) and the Cabal library (which is in the Cabal
 package).
 
-#### Install cabal-install 1.24.0.0
+#### Install cabal-install 1.24.0.0/2.0.0.1
```
erikd: Actually can you update that to use 1.24.0.2 instead of 1.24.0.0?

docs/installing-cabal.md
```diff
@@ -5,16 +5,17 @@ Cabal comes in two parts, the `cabal` executable (which is in the
...
 These instructions copy the `cabal` executable to `$HOME/bin` which is
 assumed to be on your `$PATH`. If you have an alternative location where
 you prefer to put executables, then use that instead.
 
 ```sh
-$ wget http://hackage.haskell.org/package/cabal-install-1.24.0.0/cabal-install-1.24.0.0.tar.gz
-$ tar xf cabal-install-1.24.0.0.tar.gz
-$ cd cabal-install-1.24.0.0
+$ VER=1.24.0.0 # or 2.0.0.1 if using 8.2.2 or higher.
+$ wget http://hackage.haskell.org/package/cabal-install-${VER}/cabal-install-${VER}.tar.gz
```
erikd: And use HTTPS there? (Making sure it works of course).

docs/installing-cabal.md
justanotherdot: Should I replace all instances of `1.24.0.0` to `1.24.0.2` in this doc?

erikd: Yeah, that would be a good idea.

erikd:
`$HOME/.ghc` is something GHC creates automatically whereas `$HOME/haskell` is something specific to this document.

justanotherdot:
> $HOME/.ghc is something GHC creates automatically whereas $HOME/haskell is something specific to this document.

Ah right, my bad, was mixing up the two.

docs/installing-cabal.md
justanotherdot: Actually, is it `1.24.2.0`?

justanotherdot:
`https` works, just confirming what Cabal lib comes with `1.24.0.2` so the right output is shown.

docs/installing-cabal.md
```diff
@@ -36,18 +37,18 @@ didn't use the `bootstrap.sh` script, then make sure you run `cabal
 --version` to double check the version of the Cabal library that you
 need.
 
-*GHC 8.0.1 ships with Cabal-1.24.0.0 out of the box, so you may already
+*GHC 8.0.1 ships with Cabal-1.24.0.2 out of the box, so you may already
```
erikd: I'd actually drop that section altogether. People outside Ambiata who are using GHC 8.0 should be using 8.0.2 anyway and people really should be using ghc 8.2.2 for now and starting to use 8.4.1.

erikd:
Sorry, I meant remove the bit that talks about ghc 8.0.1, not the whole thing. Should have been clearer,

justanotherdot:
> Sorry, I meant remove the bit that talks about ghc 8.0.1, not the whole thing. Should have been clearer,

Ah, got it. Too easy.

erikd (approved):

erikd:
I'll merge that when it passes CI.

PR #219, opened 2018-04-11
Support GHC 8.4.1

tmcgilchrist:


test/Test/Mafia/IO.hs
```diff
@@ -12,10 +11,7 @@ import           Test.QuickCheck.Monadic
 import           System.CPUTime (getCPUTime)
 
 testIO :: Testable a => IO a -> Property
-testIO = testPropertyIO . run
-
-testPropertyIO :: Testable a => PropertyM IO a -> Property
-testPropertyIO = monadicIO . (=<<) stop
+testIO = monadicIO . run
```
tmcgilchrist: Not sure this is doing the right thing, bumped the QuickCheck version min to 2.10 and it seems to work locally.

erikd: That seems reasonably sane.

Did you force push this branch because it doesn't seem to have my commits?

thumphries: Can probably avoid bumping the min version by writing

```haskell
testIO k = monadicIO $ do
  p <- run k
  stop p
  return p
```

tmcgilchrist: @erikd this is based off my fork and branch. Where did you make your changes?

@thumphries thanks :-)

erikd: I pushed my changes to the `topic/ghc_next` branch here on git hub. I've still got my commits so not big deal.

erikd (changes requested):
This successful build result is not reliable.

In order to avoid Travis build timouts, we install a hoogle binary from Ubuntu and hence avoid building hoogle during the tests.

I just tried this locally (where I do not have hoogle installed) and it failed building hoogle (due to semigroup/monoid changes).

Issue #198, opened 2018-01-01
Hoogle 5.x support


charleso:
Mafia currently uses the `combine` command from hoogle 4.x which is no longer supported in hoogle 5.x.

The solution appears to be to use the [generate](https://github.com/ndmitchell/hoogle/blob/14744b833b0aaf079560065fb031ee7a5860ab63/src/Action/Generate.hs#L65) command with multiple sources

```
hoogle generate --source=local --source=stackage --source=hackage --source=file1.txt 
```

Unfortunately there doesn't appear to be a common command that works between both versions.

I also suggest when upgrading to support an override hoogle on the `$PATH`, and then switch behaviour depending on the version.

erikd:
The use of Hoogle 4.x is also causing the tests suite to fail when building with ghc 8.4 because Hoogle 4.x doesn't build with ghc 8.4.

I think I've used the `mafia hoogle` command once (to try it out) in all the time I've used mafia. I actually prefer the web version of hoogle. Should we simply drop the `hoogle` sub command?

Update: @markhibberd says he uses the `hoogle` subcommand every day, so we need to keep it.

etorreborre:
As a data point I also use `hoogle` a lot more on the command line (inside ghci) than on the website.

erikd:
Ok, heard you loud and clear :).

tmcgilchrist:
Is the fix to switch to hoogle 5.* only or support both?

erikd:
I'd be in favor of only supporting 5.x but I'd be interested in hearing the views of others.

tmcgilchrist:
Looking at this now.

tmcgilchrist:
Added support for using the hoogle found on users $PATH, supports both Hoogle 4 and 5.
The basics of search and starting up a local server work. There are probably some rough edges, especially switching between 4 and 5. Please report and submit PRs if you find any. Cheers.

PR #220, opened 2018-04-20
Bump version bounds to give a build plan on GHC 8.4

gwils:
Note that this is the minimal set of bound updates that give 8.4 support -- there are other dependencies that could be upgraded too, but they are unrelated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
  - GHCVER=7.10.3 CABALVER=2.0
  - GHCVER=8.0.2 CABALVER=2.0
  - GHCVER=8.2.2 CABALVER=2.0
+ - GHCVER=8.4.1 CABALVER=2.0
 
 before_install:
  - sudo add-apt-repository -y ppa:hvr/ghc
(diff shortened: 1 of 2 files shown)


erikd:
Can't merge this at the moment.

Unfortunately just building in travis is insufficient. Even though it builds here, this can still fail to build in an environment which does not have `hoogle 4.x` installed.

The travis build currently installs `hoogle 4.x` from hvr's Ubunutu PPA and then builds/runs the test suite and all is fine. However, if `hoogle 4.x` is not installed, the test suite will try to build `hoogle 5.x` which `mafia` does not support (see #198). The reason we install hoogle rather than building the right version is that building hoogle as part of the test suite times out in travis. Furthermore, `hoogle 4.x` cannot be built with ghc 8.4 (monoid/semigroup issue from memory).

gwils:
Ok, I've read that thread now. Thanks for the explanation and link :)

tmcgilchrist:
Covered in the linked PR. Thanks @gwils

Issue #178, opened 2017-07-12
Mafia should print the version of Cabal it is built against


erikd-ambiata:
Useful for debugging.

markhibberd:
It already does, I missed this when this was raised, all the dependencies are printed via dependency info flag.

erikd:
Yes, but `mafia --version` says nothing about the cabal version. It currently does:
```
> mafia --version
mafia: 0.0.1-20180428102946-5e704ab
```
I'd like to see something like:
```
> mafia --version
mafia: 0.0.1-20180428102946-5e704ab
built with cabal version: X.Y.Z
```

tmcgilchrist:
This is now supported.

```shell
$ mafia --version
mafia: 0.0.1-20181025175327-6257dc5
built with cabal version: 2.0.1.0
```

PR #222, opened 2018-05-03
Added cabal version info to the version flag

ggranberry:
I wrote up a quick patch for #178 by just calling for the cabal version directly in Applicative. I did notice that most of the other version info is being put inside of BuildInfo_mafia when it's generated, so I can move the cabal version call over there if you think that's a better place.

### DIFF
--- a/src/Mafia/Options/Applicative.hs
+++ b/src/Mafia/Options/Applicative.hs
@@ -23,13 +23,10 @@ import qualified Data.Attoparsec.Text as A
 import           Data.String (String)
 import qualified Data.Text as T
 
-import           Distribution.Version (Version, versionNumbers)
-
 import           Options.Applicative as X
 import           Options.Applicative.Types as X
 
 import           Mafia.P
-import           Mafia.Cabal.Version
 
 import           System.Exit (exitSuccess, ExitCode(..), exitWith)
 import           System.IO (IO, putStrLn, print, hPutStrLn, stderr)
@@ -44,9 +41,6 @@ data SafeCommand a =
   | DependencyCommand
   | RunCommand RunType a
   deriving (Eq, Show)
-
-showVersion :: Version -> String
-showVersion = intercalate "." . fmap show . versionNumbers
 
 -- | Turn a parser into a ReadM
 eitherTextReader :: (e -> Text) -> (Text -> Either e a) -> ReadM a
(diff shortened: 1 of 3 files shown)


src/Mafia/Options/Applicative.hs
```diff
@@ -46,6 +49,9 @@ data SafeCommand a =
   | RunCommand RunType a
   deriving (Eq, Show)
 
+showVersion :: Version -> String
+showVersion = intercalate "." . fmap show . versionNumbers
```
novemberkilo: Just curious whether you chose this implementation of `showVersion` over https://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Version.html#v:showVersion and if so, why?

novemberkilo:
Thanks for the PR :) 

IMO this is better placed in `Setup.hs` where `BuildInfo_mafia` is being generated but please wait for comments from others before making any changes.

ggranberry:
Hey Navin, thanks for taking a look!

For the version question, I went poking around hackage and saw that the Version package that I thought we were using was deprecated: 
https://www.haskell.org/cabal/release/latest/doc/API/Cabal/Distribution-Version.html#g:2

So I ended up poking around the codebase and saw a quick function someone had already written up in Setup.hs. Just copying that definitely wasn't the best idea for code reusability, but I figured I'd get an opinion on the pull request. I guess I should have used hoogle to find the Data.Version method or used prettyShow. I'm happy to make any/all of those adjustments.

src/Mafia/Options/Applicative.hs
erikd: The file `src/Mafia/Package.hs` has `:renderVersion :: Version -> Text` :). The "render" part of the name was a convention in the place where I worked with @novemberkilo

erikd:
Pretty close. If you just replace your `showVersion` with the `renderVersion` that we already have this should be ready to go.

erikd:
Ah hang on, what you are currently printing out is the version of the `cabal` executable. What we really need here is the version of the `Cabal` library that mafia is built against (which may not be the same).

ggranberry:
Ah, good catch! I'll get that fixed up. How do you feel about having that info retrieved in Setup.hs and baked into BuildInfo_mafia?

erikd:
> How do you feel about having that info retrieved in Setup.hs and baked into BuildInfo_mafia?

If that's the way it gets the mafia version, then I'm 100% fine with it. Basically it should be getting the `Cabal` version in a similar way to how it gets the mafia version.

erikd (approved):
LGTM! Will give it a couple of days in case anyone else wants to weigh in.

erikd:
@ggranberry Sorry, i just notice that another PR was merged and we now have merge conflict on this one. Would you be able to rebase this?

ggranberry:
Yep, no problem. Went ahead and rebased and got everything merged.

erikd:
Thanks @ggranberry

novemberkilo:
Possibly a little late to the party now but this is what I see from this merge:

```
[navin:~/src/mafia] master ± ./dist/build/mafia/mafia --version
mafia: 0.0.1-20180513213719-6d591d3
built with cabal version: 2.0.0.0
[navin:~/src/mafia] master ± cabal --version
cabal-install version 2.0.0.0
compiled using version 2.0.1.0 of the Cabal library
```
Going from what @erikd mentioned above:

> Ah hang on, what you are currently printing out is the version of the cabal executable. What we really need here is the version of the Cabal library that mafia is built against (which may not be the same).

isn't this printing the wrong thing?

erikd:
@novemberkilo Yes, you are right. I will fix this.

erikd:
```
> dist/build/mafia/mafia --version
mafia: 0.0.1-20180514074447-519bd7a-M
built with cabal version: 2.0.1.1

> cabal list Cabal --installed
* Cabal
    Synopsis: A framework for packaging Haskell software
    Default available version: 2.2.0.1
    Installed versions: 2.0.1.0, 2.0.1.1
    Homepage: http://www.haskell.org/cabal/
    License:  BSD3
```
That seems a little better.

ggranberry:
My bad on this one. Wasn't aware that cabal_macros existed. Thanks for double checking @novemberkilo and thanks for patching it up for me @erikd!

PR #225, opened 2018-05-16
Remove mafia lock files.

tmcgilchrist:
Are these still useful to anyone?

cc @erikd @markhibberd


erikd:
I don't need them. Mark?

erikd (approved):

PR #227, opened 2018-06-07
Add dist-newstyle paths to submodule filter

erikd:
This just allows `mafia` to work in a git checkout where `cabal new-build`
is also being used.
    
Cabal's `new-build` set of commands store things in a top level
`dist-newstyle/` directory (similar to the `.cabal-sandbox` dirwectory)
which is likely to include cabal files.


justanotherdot (approved):

tmcgilchrist (approved):
LGTM

PR #229, opened 2018-07-07
Update installing-cabal.md

nhibberd:


justanotherdot (approved):

erikd:
Any reason for sticking to such an old version of cabal-install? Both 1.26.* and 2.0.* should work just as well. What am I missing?

I am using mafia will cabal-install 2.0 exclusively, but iI also have 2.2 installed so I can mess around with the "new-build" commands.

justanotherdot:
Actually, do we even need this section on installing the Cabal lib anymore? The link in the section is a dead one (unless we want to reintroduce it).

tmcgilchrist:
I'm ok with updating the links to 1.26 if that works.

tmcgilchrist (approved):
I'm ok with this or pointing to 1.26.

erikd:
So why not update the links so its 2.0 instead of 1.26?

tmcgilchrist:
Will add a section to do cabal 2.0 install as well.

PR #233, opened 2018-09-12
Add BSD3 as the license

tmcgilchrist:
Match the LICENSE file, it should all be under BSD3.

Also add me as the maintainer, happy to add others if they want to be actively involved.


erikd (approved):
Approved after you add me as a maintainer :)

PR #236, opened 2018-10-22
Hoogle 5 support

tmcgilchrist:


main/mafia.hs
```diff
@@ -750,7 +750,7 @@ appendStandardDirs dir =
 
 standardSourceDirs :: [Directory]
 standardSourceDirs =
-  ["src", "test", "gen", "dist/build", "dist/build/autogen"]
+  ["src", "test", "gen", "dist/build", "dist/build/autogen", "dist/build/global-autogen"]
```
erikd, line 753: Is that something new in cabal 2.0 ?

src/Mafia/Git.hs
```diff
@@ -91,7 +91,7 @@ getSubmodules = do
   root    <- getProjectRoot
   Out out <- callFrom GitProcessError root "git" ["submodule"]
 
-  sequence . fmap parseSubmoduleLine . T.lines $ out
+  traverse parseSubmoduleLine . T.lines $ out
```
erikd, line 94: :)

main/mafia.hs
tmcgilchrist: New in cabal 2.0 and GHC 8.2 apparently.

erikd (approved):
When CI is happy with this then so am I.

tmcgilchrist:
The CI hooks for travis look like they are still borked, see https://travis-ci.org/haskell-mafia/mafia/builds/444837993

domdere (commented):
LGTM, 👍 on 🤢

